PR #37, opened 2016-03-22
Set up continuous integration

dalg24:
This is a work in progress. Please DO NOT MERGE YET.


dalg24:
test

dalg24:
Ready for your review

dalg24:
``DataTransferKitOperators_SplineInterpolation_test_MPI_4`` is failing

dalg24:
I recommend merging this commit and deal with the test failure separately

dalg24:
CDash server is [here](http://jupyterdocker.ornl.gov/CDash/index.php?project=DataTransferKit)

sslattery:
One question here:

How does docker/configure_trilinos.sh get updated if we change which Trilinos packages we are using? Do we have to do this manually? If so do we see this in a pull request I guess and the build will fail if the file hasn't been updated?

dalg24:
this all have to be updated manually
note that it involves updating the docker image dtk-stk which provides the pre-built environment for the regression

PR #42, opened 2016-04-12
Adding isLocallyOwned to Entity interface.

sslattery:
Entities will need to implement this function instead of getOwnerRank()
to help specify parallel ownership. Not implemented yet for the adapters.


dalg24:
Do you want to indicate that ``ownerRank()`` will be soon deprecated?

PR #48, opened 2016-04-20
Point cloud eti

sslattery:
This adds explicit instantiation to point cloud operators and fixes some ETI build issues with tests.


packages/Operators/src/PointCloud/DTK_EuclideanDistance_impl.hpp
```diff
@@ -43,6 +43,8 @@
 
 #include <cmath>
 
+#include "DTK_EuclideanDistance.hpp"
+
```
Rombur: You forgot to remove the include in ```DTK_EuclideanDistance.hpp```

Rombur:
It's good fix the comment and merge yourself.

sslattery:
Fixed Euclidean distance. Will merge if build is clean (except possible for spline interpolation test).

PR #49, opened 2016-04-20
Fix Spline Interpolation Test

sslattery:
Adding a random number seed and loosening the tolerance. I think loosening the tolerance here is OK. The results are a function of the linear solve used within the spline operator and are therefore subject to that tolerance. This changed from 1e-7 to 1e-6.


Rombur:
The test failed because Travis had to restart :-( but you can restart it manually to see if the tests pass.

PR #51, opened 2016-04-21
Adding option to keep range data for missed points instead of zero to consistent interpolation op

sslattery:
This option lets users keep the part of the range vector that has
missing points when enabled. This automatically adds tracking of
missed range entities as well as this is needed for making the vector.
To enable this in a general way the map operator now always pulls
data from the range vector application before applying the implementation
operator. We will want to review the performance of this later.

### DIFF
--- a/packages/Operators/test/tstConsistentInterpolationOperator.cpp
+++ b/packages/Operators/test/tstConsistentInterpolationOperator.cpp
@@ -548,8 +548,8 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, many_to_many_test )
 	TEST_EQUALITY( test_val, point_dofs[i] );
     }
 
-    // Check that proc zero had some points not found.
-    int num_missed = (comm_rank != comm_size-1) ? 0 : 5;
+    // Check that the last proc had some points not found.
+    int num_missed = (comm_rank == comm_size-1) ? 5 : 0;
     Teuchos::Array<EntityId> missed_ids(
 	map_op->getMissedRangeEntityIds() );
     TEST_EQUALITY( missed_ids.size(), num_missed );
@@ -880,7 +880,7 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, keep_range_data_test )
 
     // DOMAIN SETUP
     // Don't put domain entities on proc 0.
-    int num_boxes = (comm->getRank() != 0) ? 5 : 0;
+    int num_boxes = (comm_rank != 0) ? 5 : 0;
     Teuchos::Array<DataTransferKit::SupportId> box_ids( num_boxes );
     Teuchos::ArrayRCP<double> box_dofs( num_boxes );
     Teuchos::Array<Entity> boxes( num_boxes );
@@ -956,12 +956,12 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, keep_range_data_test )
     // Check the results of the mapping.
     for ( int i = 0; i < num_points; ++i )
     {
-	double test_val = (comm_rank != comm_size-1) ? 2.0*point_ids[i] : 2.2;
+	double test_val = (comm_rank == comm_size-1) ? 2.2 : 2.0*point_ids[i];
 	TEST_EQUALITY( test_val, point_dofs[i] );
     }
 
-    // Check that proc zero had all points not found.
-    int num_missed = (comm_rank != comm_size-1) ? 0 : 5;
+    // Check that the last proc had all points not found.
+    int num_missed = (comm_rank == comm_size-1) ? 5 : 0;
     Teuchos::Array<EntityId> missed_ids(
 	map_op->getMissedRangeEntityIds() );
     TEST_EQUALITY( missed_ids.size(), num_missed );
--- a/packages/Operators/test/tstConsistentInterpolationOperator.cpp
+++ b/packages/Operators/test/tstConsistentInterpolationOperator.cpp
@@ -302,7 +302,7 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, no_domain_0_test )
     for ( int i = 0; i < num_points; ++i )
     {
 	point_ids[i] = num_points*comm_rank + i;
-	point_dofs[i] = 0.0;
+	point_dofs[i] = 2.2;
 	point[0] = 0.5;
 	point[1] = 0.5;
 	point[2] = point_ids[i] + 0.5;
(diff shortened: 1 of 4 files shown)


sslattery:
Damien can you review?

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -302,7 +302,7 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, no_domain_0_test )
     for ( int i = 0; i < num_points; ++i )
     {
 	point_ids[i] = num_points*comm_rank + i;
-	point_dofs[i] = 0.0;
+	point_dofs[i] = 2.2;
```
dalg24: why did you change that one?

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+
+    // Get the communicator.
+    Teuchos::RCP<const Teuchos::Comm<int> > comm =
+	Teuchos::DefaultComm<int>::getComm();
+    int comm_size = comm->getSize();
+    int comm_rank = comm->getRank();
+
+    // DOMAIN SETUP
+    // Don't put domain entities on proc 0.
+    int num_boxes = (comm->getRank() != 0) ? 5 : 0;
```
dalg24: use ``comm_rank`` here

packages/Operators/test/tstConsistentInterpolationOperator.cpp
sslattery: It was to make sure we would overwrite the data with zeros if we missed points. The test wasn't checking that before.

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+    Teuchos::Array<DataTransferKit::SupportId> box_ids( num_boxes );
+    Teuchos::ArrayRCP<double> box_dofs( num_boxes );
+    Teuchos::Array<Entity> boxes( num_boxes );
+    for ( int i = 0; i < num_boxes; ++i )
+    {
+	box_ids[i] = num_boxes*(comm_size-comm_rank-1) + i;
+	box_dofs[i] = 2.0*box_ids[i];
+	boxes[i] = BoxGeometry(box_ids[i],comm_rank,box_ids[i],
+		       0.0,0.0,box_ids[i],1.0,1.0,box_ids[i]+1.0);
```
dalg24: So basically boxes 1x1x1 stacked on top of each other in z-direction, right?

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+    Teuchos::Array<Entity> points( num_points );
+    for ( int i = 0; i < num_points; ++i )
+    {
+	point_ids[i] = num_points*comm_rank + i;
+	point_dofs[i] = 2.2;
+	point[0] = 0.5;
+	point[1] = 0.5;
+	point[2] = point_ids[i] + 0.5;
+	points[i] = Point(point_ids[i],comm_rank,point);
+    }
```
dalg24: 5 points per proc, located in the center of boxes of proc+1, except the last proc that will be missed

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+    map_op->setup( 
+	domain_manager.functionSpace(), range_manager.functionSpace() );
+
+    // Apply the map.
+    map_op->apply( *domain_dofs, *range_dofs );
+
+    // Check the results of the mapping.
+    for ( int i = 0; i < num_points; ++i )
+    {
+	double test_val = (comm_rank != comm_size-1) ? 2.0*point_ids[i] : 2.2;
```
dalg24: ``point_ids[i]`` just happen to be equal to the id of the box it was found in.
That does the work but man it's not easy to read...

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+	TEST_EQUALITY( test_val, point_dofs[i] );
+    }
+
+    // Check that proc zero had all points not found.
```
dalg24: no actually the proc with the highest rank is the one that has the points out of the boxes

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+    int num_missed = (comm_rank != comm_size-1) ? 0 : 5;
```
dalg24: seriously, a negation in a ternary operator?

packages/Operators/test/tstConsistentInterpolationOperator.cpp
```diff
@@ -868,5 +868,111 @@ TEUCHOS_UNIT_TEST( ConsistentInterpolationOperator, local_missed_range_test )
...
+    Teuchos::Array<EntityId> missed_ids(
+	map_op->getMissedRangeEntityIds() );
+    TEST_EQUALITY( missed_ids.size(), num_missed );
+    std::sort( point_ids.begin(), point_ids.end() );
+    std::sort( missed_ids.begin(), missed_ids.end() );
+    for ( int i = 0; i < num_missed; ++i )
+    {
+	TEST_EQUALITY( missed_ids[i], point_ids[i] );
```
dalg24: don't you want to check the dofs/values rather than the ids?

dalg24:
Here are a few other comments:
* We need to document more clearly what's going on here: If ``"Keep Missed Range Data"`` is present in the list of parameters we will leave the corresponding dofs/values UNCHANGED in the target vector.
* Do we want to allow for a user specified value to be set in the missed entities? Or is the user responsible for setting the target vector to that value before he calls apply? In that case we can add a comment to explain how that can be done.

dalg24:
NOTE: I referenced the PR in the other issue by mistake. I removed the ref but it is still showing up on the PR. I'm not sure how to remove that.

packages/Operators/test/tstConsistentInterpolationOperator.cpp
dalg24: ok

packages/Operators/test/tstConsistentInterpolationOperator.cpp
sslattery: Here we are making sure that the missing ids match the point ids - not the values.

dalg24: OK you check values just before

PR #54, opened 2016-04-27
Node to node operator

sslattery:
Adding a basic node-to-node transfer map for aligned surface meshes in the point cloud operators.


sslattery:
Damien can you review?

dalg24:
Will do

packages/Operators/src/Factory/DTK_PointCloudOperatorFactory.hpp
```diff
@@ -90,7 +90,8 @@ class PointCloudOperatorFactory
     enum PointCloudOperatorType
     {
 	SPLINE_INTERPOLATION,
-	MOVING_LEAST_SQUARE
+	MOVING_LEAST_SQUARE,
+        NODE_TO_NODE
     };
```
dalg24: what's the point of the ``enum``s here? why don't you use directly the ``std::string`` as the key for the factory?

dalg24: also it might be a good idea to use a static map (``std::string`` to builder) and use register in derived classes.
the switch statement is hard to maintain and you have to recompile each time you add a new operator.

dalg24: these are only suggestions for improvement and don't need to be addressed in this PR.

sslattery: This type of factory design I had seen in Trilinos and emulated here.

I agree a builder/factory pattern would be more maintainable. That would also let users register their own stuff way easier (a different basis type for example). I think we should make an issue to refactor our factory scheme.

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+	  ++domain_entity, ++entity_counter )
+    {
+	domain_space->shapeFunction()->entitySupportIds(
+	    *domain_entity, source_node_supports );
+	DTK_CHECK( 1 == source_node_supports.size() );
+	source_support_ids[entity_counter] = source_node_supports[0];
+	domain_space->localMap()->centroid(
+	    *domain_entity, source_centers(DIM*entity_counter,DIM) );
+    }
+
```
dalg24: you do the exact same thing for source and target.
it calls for a method that takes a function space and returns nodes coord and id.

sslattery: you are right - there should be a function that does that. now that i think about it all of the point cloud operators do this so i should clean that up.

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+    // Take the proximity to be 0.001 of the largest distance.
+    target_proximity *= 1.0e-3;
+    
+    // Gather the source centers that are in the proximity of the target
+    // centers on this proc.
+    Teuchos::Array<double> dist_sources;
+    CenterDistributor<DIM> distributor( 
+	comm, source_centers(), target_centers(), target_proximity, dist_sources );
```
dalg24: Remind me where the kd-tree is built. It sounds silly to me that we use the proximity if the tree already exists.
In that case I would just do a ``nnSearch`` for k=1.

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+                < 1.0e-14 );
+                
+            // Get the id of the domain node
+            indices[0] =
+                dist_source_support_ids[ pairings.childCenterIds(i)[0] ];
+            
+	    // Populate the coupling matrix row.            
+	    d_coupling_matrix->insertGlobalValues( 
+		target_support_ids[i], indices(), values() );
+	}
```
dalg24: So it means we allow for target nodes that are not found.

packages/Operators/test/tstNodeToNodeOperator.cpp
```diff
@@ -0,0 +1,196 @@
...
+    setupAndRunTest(
+	"node_to_node_test.xml", gold_data, test_result );
+    
+    // Check the results.
+    TEST_EQUALITY( gold_data.size(), test_result.size() );
+    int num_points = gold_data.size();
+    for ( int i = 0; i < num_points; ++i )
+    {
+	TEST_FLOATING_EQUALITY( gold_data[i], test_result[i], epsilon );
+    }
```
dalg24: did you check or ask whether these unit testing macros from Teuchos can be called inside a function?

packages/Operators/test/tstNodeToNodeOperator.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include "Teuchos_DefaultComm.hpp"
+#include "Teuchos_CommHelpers.hpp"
+#include "Teuchos_ParameterList.hpp"
+#include "Teuchos_XMLParameterListCoreHelpers.hpp"
+
+//---------------------------------------------------------------------------//
+// Test epsilon.
+//---------------------------------------------------------------------------//
+
+const double epsilon = 1.0e-8;
```
dalg24: a global variable? booooo

dalg24: that sounds VERY relaxed for this tolerance

dalg24: I just checked and it seems the macro is using a relative tolerance

dalg24: I suspect subtracting the two values and making sure the absolute value is less than say 1e-14 is a better check

dalg24: what do you think?

sslattery: I agree this check should be tightened in the node to node map - the results should be exact. I will change this.

sslattery: by the way i like the global variable in the unit tests here. we never worry about scope issues and we can use the same epsilon in all of the tests. also based on its location and the comment bar its really obvious where it is.

dalg24: fair enough for ``epsilon``

dalg24:
Can one of the admins verify this patch?

dalg24:
test this please

dalg24:
ok to test

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
sslattery: Yeah I should probably just expand the box by a small amount (like 1.0e-3) to make sure we get the points within a floating point tolerance.

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
sslattery: Right - in Srikanth's case he was originally doing a node-to-node map with overlap so some points didn't get data.

sslattery:
Code has been updated per the review. Damien can you take a look when you get a chance?

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
```diff
@@ -239,6 +197,35 @@ void NodeToNodeOperator<DIM>::applyImpl(
 }
 
 //---------------------------------------------------------------------------//
+// Extract node coordinates and ids from an iterator.
+template<int DIM>
+void NodeToNodeOperator<DIM>::getNodeCoordsAndIds(
+    const Teuchos::RCP<FunctionSpace>& space,    
+    EntityIterator iterator,
```
dalg24: can't you get ``iterator`` from ``space``?

packages/Operators/src/PointCloud/DTK_NodeToNodeOperator_impl.hpp
```diff
@@ -103,26 +103,14 @@ void NodeToNodeOperator<DIM>::setupImpl(
 	domain_iterator =
 	    domain_space->entitySet()->entityIterator( 0, domain_predicate );
     }
```
dalg24: couldn't that block that gets the iterator move into the ``getNodeCoordsAndIds`` function?

sslattery: in this case we can move that code inside

sslattery:
OK this should be the last change @dalg24

PR #56, opened 2016-04-27
[WIP] C/Fortran API

dalg24:


dalg24:
Can one of the admins verify this patch?

packages/Adapters/POD_C/src/DTK_NoName.h
```diff
@@ -0,0 +1,48 @@
...
+
+// Opaque pointer
+typedef void DTK_Map;
+
+// Possible data layouts:
+// * Contiguous 
+//     X1 X2 ... Xn Y1 Y2 ... Yn
+// * Interleave 
+//     X1 Y1 X2 Y2 ... Xn Yn
+typedef enum data_layout { CONTIGUOUS, INTERLEAVE } DTK_DataLayout;
```
sslattery: Should we make these DTK_CONTIGUOUS and DTK_INTERLEAVE to avoid collisions or does the enum prevent this?

packages/Adapters/POD_C/src/DTK_NoName.h
```diff
@@ -0,0 +1,48 @@
...
+//----------------------------------------------------------------------------//
+// Create a DTK map operator
+DTK_Map * create_map( MPI_Comm comm,
+                      double const * src_coord,
+                      DTK_DataLayout src_layout,
+                      double const * tgt_coord,
+                      DTK_DataLayout tgt_layout,
+                      int space_dim,
+                      char const * options );
+
```
sslattery: Will we let users give us null pointers? Do we expect allocated pointers of size zero?

packages/Adapters/POD_C/src/DTK_NoName.h
```diff
@@ -0,0 +1,48 @@
...
+// * Contiguous 
+//     X1 X2 ... Xn Y1 Y2 ... Yn
+// * Interleave 
+//     X1 Y1 X2 Y2 ... Xn Yn
+typedef enum data_layout { CONTIGUOUS, INTERLEAVE } DTK_DataLayout;
+
+//----------------------------------------------------------------------------//
+// Create a DTK map operator
+DTK_Map * create_map( MPI_Comm comm,
+                      double const * src_coord,
```
sslattery: How do you know how many source and target points there are in this interface?

packages/Adapters/POD_C/src/DTK_NoName.h
```diff
@@ -0,0 +1,48 @@
...
+                      DTK_DataLayout src_layout,
+                      double const * tgt_coord,
+                      DTK_DataLayout tgt_layout,
+                      int space_dim,
+                      char const * options );
+
+//----------------------------------------------------------------------------//
+// Apply the map
+void apply_map( DTK_Map * dtk_map,
+                double const * src_field,
```
sslattery: Based on the field dimension and the number of source and target points we should know how much field data is here. Should we also make the users tell us so we can check?

packages/Adapters/POD_C/src/DTK_NoName.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+
+//----------------------------------------------------------------------------//
+void apply_map( DTK_Map * dtk_map,
+                double const * src_field,
+                DTK_DataLayout src_layout,
+                double * tgt_field,
+                DTK_DataLayout tgt_layout,
+                int dim,
```
sslattery: maybe call this variable field_dim instead of dim

dalg24:
Refer to this link for build results (access rights to CI server needed): 
/job/DataTransferKit/15/

dalg24:
Refer to this link for build results (access rights to CI server needed): 
/job/DataTransferKit/16/

dalg24:
Refer to this link for build results (access rights to CI server needed): 
/job/DataTransferKit/18/

packages/Adapters/POD_C/src/DTK_NoName.h
dalg24: if size is zero it does not matter what the pointer points to.

packages/Adapters/POD_C/src/DTK_NoName.h
dalg24: I'm not sure if we should ask. What do you think?

dalg24:
Refer to this link for build results (access rights to CI server needed): 
/job/DataTransferKit/19/

dalg24:
Refer to this link for build results (access rights to CI server needed): 
/job/DataTransferKit/20/

packages/Adapters/POD_C/src/DTK_C_API.h
```diff
@@ -0,0 +1,49 @@
...
+
+// Convenience typedef for an opaque pointer
+typedef void DTK_Map;
+
+// Possible data layouts:
+// * Contiguous 
+//     X1 X2 ... Xn Y1 Y2 ... Yn
+// * Interleave 
+//     X1 Y1 X2 Y2 ... Xn Yn
+typedef enum data_layout { DTK_CONTIGUOUS, DTK_INTERLEAVE } DTK_Data_layout;
```
sslattery: Instead of DTK_CONTIGUOUS let's use DTK_BLOCKED. The memory is always contiguous here and blocked represents that the data is blocked by dimension instead of interleaved by dimension.

packages/Adapters/POD_C/test/tstC_API.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+  // let us use INFO format
+  std::string options;
+  options = "\n"\
+    "\"Map Type\" \"Spline Interpolation\"\n"\
+    "\"Basis Type\" \"Wendland\"\n"\
+    "\"Basis Order\" 0\n"\
+    "\"Spatial Dimension\" 3\n"\
+    "\"Target Coord Data Layout\" \"Contiguous\"\n"\
+    "\"Source Coord Data Layout\" \"Contiguous\"\n"\
```
sslattery: Why do dimension and layout need to be in options if they are input parameters in the API?

packages/Adapters/POD_C/test/tstC_API.cpp
```diff
@@ -0,0 +1,92 @@
...
+                                     tgt_num,
+                                     DTK_CONTIGUOUS,
+                                     space_dim,
+                                     options.c_str() );
+
+  options = "\n"\
+    "\"Source Field Data Layout\" \"Interleave\"\n"\
+    "\"Target Field Data Layout\" \"Interleave\"\n"\
+    "\"Field Dimension\" 1\n"\
+    "\n";
```
sslattery: Same comment here on options. What options might we need on apply? Do we need any at all?

packages/Adapters/POD_C/src/DTK_C_API.h
```diff
@@ -0,0 +1,49 @@
...
+                         char const*     options );
+
+//----------------------------------------------------------------------------//
+void DTK_Map_apply( DTK_Map*        dtk_map,
+                    double const*   src_field,
+                    DTK_Data_layout src_layout,
+                    double*         tgt_field,
+                    DTK_Data_layout tgt_layout,
+                    int             field_dim,
+                    char const*     options );
```
sslattery: Not sure options are needed here. Our apply interface has not parameter list.

packages/Adapters/POD_C/src/DTK_C_API.cpp
```diff
@@ -0,0 +1,246 @@
...
+  boost::property_tree::ptree ptree;
+  boost::property_tree::read_info(ss, ptree);
+  // QUESTION: Should we go ahead and just parse all but the comm and ptrs?
+  Teuchos::ParameterList parameters;
+  parameters.set("Map Type", ptree.get<std::string>("Map Type"));
+  parameters.set("Spatial Dimension", ptree.get<int>("Spatial Dimension"));
+  parameters.set("Basis Type", ptree.get<std::string>("Basis Type"));
+  parameters.set("Basis Order", ptree.get<int>("Basis Order"));
+  // TODO: kNN
+  parameters.set("RBF Radius", ptree.get<double>("RBF Radius"));
```
sslattery: Did you look to see if there is a parameter list to property tree adapter class in Trilinos? I doubt there is but not sure. They have a boost::Shared_Ptr to Teuchos::RCP mapping so maybe.

packages/Adapters/POD_C/src/DTK_C_API.cpp
```diff
@@ -0,0 +1,246 @@
...
+    src_field, src_local_num, src_layout, field_dim );
+  DTK_CHECK( range_field.size() == src_local_num*field_dim );
+
+  auto range_map = map_operator->getRangeMap();
+  DTK_CHECK( range_map->isContiguous() );
+  unsigned tgt_local_num = range_map->getMaxLocalIndex() - range_map->getMinLocalIndex() + 1;
+  Teuchos::Array<double> domain_field(tgt_local_num*field_dim);
+
+  // TODO: I suspect I need the domain and range managers to build the
+  // vectors...
```
sslattery: We can get the range and domain maps right from the operator - no need to cache additional information here. The entity centered field is the right choice. I'll help you finish this up.

packages/Adapters/POD_C/src/DTK_C_API.h
```diff
@@ -0,0 +1,49 @@
...
+#include "mpi.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// QUESTION: Do we want to be able to track missed entities?
+
+// Convenience typedef for an opaque pointer
+typedef void DTK_Map;
```
sslattery: I think we can get away with just this typedef. I haven't seen a need for an additional struct yet.

packages/Adapters/POD_C/src/DTK_C_API.h
```diff
@@ -0,0 +1,49 @@
+#ifndef DTK_NONAME_H
+#define DTK_NONAME_H
+
+#include "mpi.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// QUESTION: Do we want to be able to track missed entities?
```
sslattery: At some point we do but the point cloud operators don't have this capability yet. We may want to add that as part of this work.

packages/Adapters/POD_C/src/DTK_C_API.cpp
```diff
@@ -0,0 +1,246 @@
...
+      entities[i] = DataTransferKit::Point(global_id, comm_rank, coord);
+    }
+  else
+    throw std::runtime_error("Invalid data layout");
+  return entities;
+}
+
+//----------------------------------------------------------------------------//
+Teuchos::Array<double>
+  pull_field(
```
sslattery: Entity centered field will handle both pull_field and push_field for interleaved and blocked data

packages/Adapters/POD_C/src/DTK_C_API.cpp
```diff
@@ -0,0 +1,246 @@
...
+                                    &local_num,
+                                    shift.size(),
+                                    shift.data() );
+  for (std::size_t k = 1; k < comm_size; ++k)
+    shift[k] = shift[k-1];
+  shift[0] = 0;
+  for (std::size_t k = 1; k < comm_size; ++k)
+    shift[k] += shift[k-1];
+
+  // TODO: might be worth implementing some data extractor to be used both here
```
sslattery: I think we want to write our own entity, iterator, and entity set implementations here for efficiency - no need to copy the data. The basic geometry adapters do not have this capability so right now we would have to do a ton of copies. This should be pretty straightforward to implement.

packages/Adapters/POD_C/src/DTK_C_API.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+  //auto domain_vector = DataTransferKit::FieldMultiVector(
+  //  Teuchos::rcp(new DataTransferKit::EntityCenteredField(domain_entities(), field_dim, domain_field)),
+  //  domain_manager.functionSpace()->entitySet() );
+
+  // Apply the map operator
+  // TODO:
+  // map_operator->apply(domain_vector, range_vector);
+
+  push_field(domain_field, tgt_field, tgt_layout, field_dim);
```
sslattery: we will get rid of this and have the operator apply do everything.

packages/Adapters/POD_C/src/DTK_C_API.h
sslattery: Does a void typedef work or does it have to be void*?

packages/Adapters/POD_C/src/DTK_C_API.h
dalg24: OK, will change that

packages/Adapters/POD_C/test/tstC_API.cpp
dalg24: It was purely illustrative to show that we could in practice have them here and reduce the number of input params.

packages/Adapters/POD_C/test/tstC_API.cpp
dalg24: We do not need them

packages/Adapters/POD_C/src/DTK_C_API.cpp
dalg24: From what I gathered by looking at the Doxygen documentation for ``Teuchos``, they only have a XML reader.

packages/Adapters/POD_C/test/tstC_API.cpp
sslattery: OK. I would like all input values related to the content and structure of the data pointers to come in through the API - that way we know whats there and tracking down set faults from bad inputs is more clear. Anything we would stick in a parameter list and pass down to a factory can come in through the options.

packages/Adapters/POD_C/src/DTK_C_API.h
dalg24: Not sure I understand the question.
I did not do
```
typedef void* DTK_Map;
```
so that user sees he is dealing with a pointer and understand he is responsible for calling delete.

packages/Adapters/POD_C/src/DTK_C_API.h
dalg24: I see.

packages/Adapters/POD_C/src/DTK_C_API.cpp
dalg24: Ok so it means I don't need to reorder the data. Good.

sslattery:
This looks good. I am going to merge and then add my updates in another PR.

PR #61, opened 2016-05-18
Implemented Fortran API

dalg24:
test does not actually assert anything
it is more an example than a test


packages/Adapters/POD/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-TRIBITS_SUBPACKAGE(C_API)
+TRIBITS_SUBPACKAGE(PlainOldData)
```
sslattery: I actually think it will be more clear if we have two packages: C_API and Fortran_API. We will likely have a lot of C and Fortran work in our future so this would help. Also it separates the libraries we will construct - one library per subpackage. Even though there is only one Fortran file for now, I think this makes sense.

packages/Adapters/POD/src/CMakeLists.txt
```diff
@@ -23,7 +21,22 @@ APPEND_SET(SOURCES
   ) 
 
 TRIBITS_ADD_LIBRARY(
-  dtk_c_api
+  dtk_pod
   HEADERS ${HEADERS}
   SOURCES ${SOURCES}
   )
+
+TRIBITS_ADD_LIBRARY(
```
sslattery: Why do we have to split up the C/C++ code? Isn't the extern "C" good enough for linking with other C codes?

packages/Adapters/POD/src/DTK_C_API.h
```diff
@@ -34,6 +34,7 @@
 #ifndef DTK_NONAME_H
```
sslattery: We should make the header protection macros match the header file name

packages/Adapters/POD/test/tstFortran_API.F90
```diff
@@ -0,0 +1,69 @@
...
+  src_coord(3) = 0
+  src_coord(4) = comm_rank + 0.5
+  src_field(2) = src_coord(4)
+  tgt_coord(1) = 0
+  tgt_coord(2) = comm_size - comm_rank - 1
+  tgt_field(1) = 255
+  test = .FALSE.
+
+  apply_transpose = .FALSE.
+  options = '{ "Map Type": "Node To Node" }'//c_null_char
```
sslattery: What is the //c_null_char for? Is this for end of line formatting?

dalg24: yes

dalg24: char array terminated by '\0' in C

packages/Adapters/POD/test/tstFortran_API.F90
```diff
@@ -0,0 +1,69 @@
...
+  
+  dtk_map = DTK_Map_create(MPI_COMM_WORLD, &
+      c_loc(src_coord), src_num, DTK_INTERLEAVED, &
+      c_loc(tgt_coord), tgt_num, DTK_BLOCKED, &
+      space_dim, options)
+
+  call DTK_Map_apply(dtk_map, c_loc(src_field), DTK_INTERLEAVED, &
```
sslattery: are c_loc and the other interoperability keywords portable? or are they only for gcc compilers?

sslattery: OK I'm seeing documentation for both intel and gcc. We will have to see how this goes on the Argonne machines.

sslattery:
I think this looks pretty good. I want to talk a little more before merging though - mostly to get your opinion of where things are at.

dalg24:
Ready for merge

dalg24:
The failing test belongs to libmesh

sslattery:
OK this looks good. We will have to see what is going on with the lib mesh test. If it shows up on the master build after this merge we will address.

PR #63, opened 2016-06-10
Fix C/Fortran API TPLs dependencies

dalg24:


dalg24:
I believe the README is fine as it is now.
The libmesh adapters also requires MPI, tribits will resolve this at configure time.

Rombur:
What about the new ```Boost``` dependency? Don't we need to update the configuration example?

dalg24:
You are right. I submitted a new PR.

PR #66, opened 2016-06-15
Add a check that entity are not found and missed at the same time.

Rombur:


Rombur:
With this change, Srikanth's test passes.

packages/Operators/src/Search/DTK_ParallelSearch.cpp
```diff
@@ -42,6 +42,7 @@
 #include "DTK_DBC.hpp"
 
 #include <Tpetra_Distributor.hpp>
+#include <Teuchos_Comm.hpp>
```
dalg24: why did you add this header?

Rombur: I was using ```allReduce``` for the check but  then I realized it was not necessary but I forgot to remove it.

dalg24:
[L254](https://github.com/Rombur/DataTransferKit/blob/358a5d22f22a905d0f1d5834af5aa98a09d13ad3/packages/Operators/src/Search/DTK_ParallelSearch.cpp#L254): No need to allocate more than ``import_missed.size()``.

packages/Operators/src/Search/DTK_ParallelSearch.cpp
```diff
@@ -250,27 +250,33 @@ void ParallelSearch::search(
...
-	// Create a unique list of missed entities without the false positives.
-	std::sort( d_missed_range_entity_ids.begin(), missed_range_end );
-	auto missed_range_unique_end = std::unique(
-	    d_missed_range_entity_ids.begin(), missed_range_end );
 	d_missed_range_entity_ids.resize(
 	    std::distance(d_missed_range_entity_ids.begin(),
-			  missed_range_unique_end) );
+			  missed_range_end) );
+
+#if HAVE_DTK_DBC
```
sslattery: This is what was causing the tests to fail (thrown exception). Is the math right here? I removed this block of code and everything passes.

Rombur: What do you mean? Before I changed the order of ```unique``` and ```set_difference``` this would throw when using 6 processors. Now it doesn't throw anymore.

Rombur:
With this change. All the tests pass on my machine.

sslattery:
ok - everything passes on my machine now with the latest commit. if travis comes back clean i will merge

Rombur:
Only Libmesh failed
 47 - DataTransferKitLibmeshAdapters_LibmeshVariableField_test_MPI_4 (Failed)

sslattery:
OK - everything passed except the known lib mesh test that fails.

PR #67, opened 2016-07-21
Set linker language for Fortran API test

dalg24:
Was not able to compile with intel compiler


sslattery:
closed by mistake - code is ready but need to investigate moab failures

PR #89, opened 2016-10-07
Remove deprecated CDash scripts

dalg24:


dalg24:
Doesn't mean we are never going to use CDash again.  Just these haven't been use in over a year and are completely outdated.

dalg24:
This is safe to merge.

PR #91, opened 2016-10-10
POD kNN

sslattery:
Adding kNN support to POD interface and fixing a bug in the POD iterator.

### DIFF
--- a/packages/Adapters/C_API/src/DTK_POD_PointCloudEntitySet.cpp
+++ b/packages/Adapters/C_API/src/DTK_POD_PointCloudEntitySet.cpp
@@ -113,15 +113,20 @@ EntityIterator POD_PointCloudEntitySet::entityIterator(
     const int topological_dimension,
     const PredicateFunction& predicate ) const
 {
-    DTK_REQUIRE( 0 == topological_dimension );
+    EntityIterator iterator;
     
-    return POD_PointCloudEntityIterator( d_cloud_coords,
-                                         d_global_ids,
-                                         d_num_points,
-                                         d_space_dim,
-                                         d_layout,
-                                         d_comm->getRank(),
-                                         predicate );
+    if ( 0 == topological_dimension )
+    {
+        iterator = POD_PointCloudEntityIterator( d_cloud_coords,
+                                                 d_global_ids,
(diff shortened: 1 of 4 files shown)


packages/Adapters/C_API/src/DTK_POD_PointCloudEntitySet.cpp
```diff
@@ -113,15 +113,20 @@ EntityIterator POD_PointCloudEntitySet::entityIterator(
...
+    if ( 0 == topological_dimension )
+    {
+        iterator = POD_PointCloudEntityIterator( d_cloud_coords,
+                                                 d_global_ids,
+                                                 d_num_points,
+                                                 d_space_dim,
+                                                 d_layout,
+                                                 d_comm->getRank(),
+                                                 predicate );
+    }
```
dalg24: when would this would be called with `topological_dimension > 0`?

sslattery: it gets called in the default bounding box implementation for entity sets

packages/Adapters/C_API/test/tstPOD_C_API.cpp
```diff
@@ -85,17 +81,61 @@ TEUCHOS_UNIT_TEST( C_API, create_apply_delete_map )
...
+  unsigned size = gold_data.size();
+  for (unsigned i = 0; i < size; ++i)
+  {
+    TEST_FLOATING_EQUALITY( gold_data[i], tgt_field[i], rel_tol );
+    TEST_COMPARE( std::abs(gold_data[i] - tgt_field[i]), <, abs_tol );
+  }
+}
+
+//---------------------------------------------------------------------------//
+TEUCHOS_UNIT_TEST( C_API, knn_map )
```
dalg24: This is the same as the previous ``TEUCHOS_UNIT_TEST ``, ``options`` is the only thing that changes.
I'd advocate for a ``for (std::string const & options : { ... }) { ... }``

sslattery: i like having separate tests so its more obvious what is failing - in this case radius search vs kin

PR #95, opened 2016-10-11
DTKData

dalg24:
Good ole symlink technique will still be good but now we have the option to pass `--recursive` to the `git clone` command.


dalg24:
This is ready for review/merge.

Rombur:
Since only one libmesh test does not pass and it doesn't have anything to do with this PR, I will merge.

Rombur:
@dalg24 When I approve the request, it means that I am fine for you to merge it if the tests pass.

dalg24:
OK

PR #154, opened 2016-12-14
Run multiple isolated environment on the same Docker host

dalg24:


docker/docker-compose.yml
```diff
@@ -10,7 +10,6 @@ services:
     # Append user name to the container name.  Specifying a container name
     # implies that there can only be one instance running at the same time on
     # the Docker daemon (per user).
-    container_name: dtk_dev_$USER
```
Rombur: So we don't give a default name to container anymore?

dalg24: the name will be <namespace>_dtk_dev_<#> where <namespace> default to "docker" (directory that contains the Compose file).

dalg24: Exactly.

PR #182, opened 2017-01-11
Making shortlog work by adding mailmap

aprokop:


Rombur:
Two questions:
 1) what does this do?
 2) why am I not in it? Is it only for the cool kids :-(

aprokop:
@Rombur `git help shortlog`
It essentially allows one to quickly see the aggregated commits by each author for instance to figure out what you worked on since past release and such. The commit solves the problem when one person committed using different names and/or emails, which will be listed separately.

 Your name is not there because you only used a single email address and a single name for the whole history of the project, thus no need for it.

dalg24:
Where does the "<username>@users.noreply.github.com" comes from?

Rombur:
@aprokop OK thanks for the explanation

aprokop:
> Where does the "<username>@users.noreply.github.com" comes from?

If in you github settings you set to hide the password, this is an email that is used when you merge PR.

dalg24:
What password?  You mean "Keep my email address private"?

aprokop:
Yeah, a misprint :) "to keep email private"

PR #190, opened 2017-03-21
Fix strings comparison in C API for "Search Type"

dalg24:
When parsing the options for the type of search we were comparing a C string using the operator==.


Rombur:
I just checked with Damien and the current code is correct http://www.cplusplus.com/reference/string/string/operators/

dalg24:
False alert...
Closing per @Rombur's comment

Issue #193, opened 2017-03-22
Generate Sphinx and Doxygen documentation as part of our CI


dalg24:
Currently we do not ensure that the doc builds without errors when we push changes.

dalg24:
- [ ] sphinx documentation (published on https://datatransferkit.readthedocs.io; updated when changes are pushed on master)
- [ ] doxygen documentation (published on gh-pages; updated manually at the moment)

dalg24:
Closing as both #194 and #195 got merged into master.  We might consider reopen if we decide to publish an updated version of the Doxygen documentation when changes are pushed to master.

PR #195, opened 2017-03-22
Use Travis CI to build doxygen documentation

dalg24:
Finish addressing #193.

This test will ensure that the Doxygen documentation builds without raising any warnings.
The bad news is that we currently have 86 of them that will need to be resolved.

We can either merge early and fix the warnings later OR wait and push fixes to this PR before we merge.  What do you all think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - TEST_SUITE=format_cpp
   - TEST_SUITE=no_trailing_whitespaces_or_tabs
   - TEST_SUITE=build_sphinx_documentation
+  - TEST_SUITE=doxygen_documentation
 
 before_install:
   - docker pull dalg24/clang-format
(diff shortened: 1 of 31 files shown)


dalg24:
The Travis CI build `doxygen_documentation` did run.  As expected, it is failing.

dalg24:
@sslattery I brought the number of warnings from 86 to 7.  Would you please take care of the few remaining ones?

Note that the Docker image is missing LaTeX at the moment.  I will add it once we get rid of all the current warnings.  Do not merge until I say it is ready.

dalg24:
Here is how to use the Docker image used on our Travis CI build to generate the doc and check for warnings.  From the DTK root dir, run:
```bash
[host]$ docker run --rm -v $PWD:/scratch -w /scratch dalg24/clang-format ./docker/travis/check_doxygen_documentation.sh
```
Consider using `<...> | grep warning:`.

dalg24:
I suggest to make this one blocking for #187.  (Which means we need would need to expedite it.)

dalg24:
Ready for review/merge

Rombur:
OK to merge once jenkins is happy

PR #234, opened 2017-06-02
[WIP] Start work on cuda-mpi.

Rombur:
This is not ready for review. I just want to make sure that tests fail the "correct" way.


Rombur:
This is so outdated that there is no point in keeping it around.

PR #251, opened 2017-06-14
Fix invalid syntax in requirements file

dalg24:
No comment.

Amend to #250


dalg24:
This time did the trick.  See [here](http://readthedocs.org/projects/datatransferkit/builds/5561231/)

PR #258, opened 2017-06-21
Add missing include.

Rombur:
I can't compile DTK without these include with gcc 7.1. `Interface_Init` tests are not passing on my machine but I am still investigating.


dalg24:
> I am still investigating.

@Rombur You want to make this a WIP then?  Or are you working on your gh statistics like Andrey?

Rombur:
@dalg24 it's an unrelated problem. This is because nobody tried to compile Kokkos in serial only...

PR #268, opened 2017-07-18
Fix warnings from clang.

Rombur:
Fix warnings from jenkins.

### DIFF
--- a/packages/Search/src/DTK_LinearBVH_decl.hpp
+++ b/packages/Search/src/DTK_LinearBVH_decl.hpp
@@ -33,7 +33,7 @@ struct TreeTraversal;
  * Bounding Volume Hierarchy.
  */
 template <typename DeviceType>
-struct BVH
+class BVH
 {
   public:
     BVH( Kokkos::View<Box const *, DeviceType> bounding_boxes );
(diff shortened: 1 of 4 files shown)


packages/Search/src/DTK_LinearBVH_decl.hpp
```diff
@@ -33,7 +33,7 @@ struct TreeTraversal;
  * Bounding Volume Hierarchy.
  */
 template <typename DeviceType>
-struct BVH
+class BVH
```
dalg24: Was about this change?

Rombur: clang complains that the explicit instantiation is for a class not for a struct

dalg24: OK, so in order to be consistent with the explicit instantiation in the _def file ([here](https://github.com/Rombur/DataTransferKit/blob/d77f255a1be53d29c919dddd48865484c94841ad/packages/Search/src/DTK_LinearBVH_def.hpp#L98)).

dalg24: class or struct does not really matter.  Please commit that change in a separate commit.
Say the 1st commit is double brace initialization and this one to be consistent with explicit instantiation.

PR #276, opened 2017-07-25
Suppress a warning when using clang (clang does not have the  flag).

Rombur:


dalg24:
See compiler warnings analysis on Jenkins [here](https://cloud.cees.ornl.gov/jenkins-ci/job/DataTransferKit-experimental/406/)

Rombur:
Forgot `Wall` on my gcc built :/

Issue #282, opened 2017-08-01
Radial Basis Function Infrastructure


sslattery:
DTK 2.0 radial basis function implementation will be ported to Kokkos device functions to allow for batched execution and composition of point cloud operators

API in 2.0 is [here](https://github.com/ORNL-CEES/DataTransferKit/blob/dtk-2.0/packages/Operators/src/PointCloud/DTK_RadialBasisPolicy.hpp)

Should implement all available orders of:

- [x] Wu Basis

- [x] Buhmann Basis

- [x] Wendland Basis

We do not need gradient evaluations at this time.

sslattery:
Will be closed with merge of #290

PR #290, opened 2017-08-08
Add compactly supported radial basis functions

dalg24:
Addressing #282


cmake/Dependencies.cmake
```diff
@@ -2,5 +2,6 @@ TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
   SUBPACKAGES_DIRS_CLASSIFICATIONS_OPTREQS
     Utils                 packages/Utils                  SS  REQUIRED
     Interface             packages/Interface              SS  REQUIRED
+    Operators             packages/Operators              SS  REQUIRED
```
sslattery: What do you think about `Maps` instead of `Operators`?

dalg24: Please ignore all changes related to creating a new subpackage and focus on the actual implementation of the CSRBFs.  We will rebase the PR this as necessary.

dalg24:
I rebased onto master and moved everything to the new "Meshfree" package.
I think it is ready for merge.

PR #296, opened 2017-08-11
Use llvm-symbolizer with Sanitizer

sslattery:
Sanitizer errors are impossible to debug without line and file info. Setting ASAN_SYMBOLIZER_PATH to the llvm-symbolizer path enables this as shown [here](https://stackoverflow.com/questions/24566416/how-do-i-get-line-numbers-in-the-debug-output-with-clangs-fsanitize-address)

I did this recently with an error and it worked great. This will always enable it with docker and clang sanitizer builds so we should also see it in our output logs in the CI system

### DIFF
--- a/scripts/docker_clang_env.sh
+++ b/scripts/docker_clang_env.sh
@@ -12,6 +12,10 @@ if [ "$SANITIZER" == "undefined_sanitizer" ]
   then FLAGS="-fsanitize=address -fsanitize=undefined -fsanitize-blacklist=/scratch/source/trilinos/release/DataTransferKit/scripts/undefined_blacklist.txt"
   # Suppress leak sanitizer on functions we have no control of.
   export LSAN_OPTIONS=suppressions=/scratch/source/trilinos/release/DataTransferKit/scripts/leak_blacklist.txt
+  # Set the ASAN_SYMBOLIZER_PATH environment variable to point to the
+  # llvm-symbolizer to make AddressSanitizer symbolize its output.  This makes
+  # the reports more human-readable.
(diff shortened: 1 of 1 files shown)


scripts/docker_clang_env.sh
```diff
@@ -12,6 +12,8 @@ if [ "$SANITIZER" == "undefined_sanitizer" ]
   then FLAGS="-fsanitize=address -fsanitize=undefined -fsanitize-blacklist=/scratch/source/trilinos/release/DataTransferKit/scripts/undefined_blacklist.txt"
   # Suppress leak sanitizer on functions we have no control of.
   export LSAN_OPTIONS=suppressions=/scratch/source/trilinos/release/DataTransferKit/scripts/leak_blacklist.txt
+  # Set the symbolizer so we get line outputs from errors for debugging.
```
dalg24: What the h does "[symbolizing the reports"](https://clang.llvm.org/docs/AddressSanitizer.html#symbolizing-the-reports)" means?

sslattery: all i know is this gives me file and line info so i can actually debug with it

sslattery: it actually parses the addresses to make it human-readable

dalg24: I'd change the comment to something like:
```bash
# Set the ASAN_SYMBOLIZER_PATH environment variable to point to the llvm-symbolizer to make AddressSanitizer symbolize its output.  This makes the reports more human-readable.
```

sslattery: done

PR #299, opened 2017-08-24
Fix bug in PointInCell

Rombur:
When FineSearch was renamed to PointInCell and move from Search to
Discretization, a bug was introduce in the CMake build. Namely, files in
Discretization where not compiled. So, we fixed the CMake bug and other bugs
found since the files are compiled now.


sslattery:
Looks like the ETI is messed up

PR #301, opened 2017-09-06
Remove useless code.

Rombur:
This code is useless. Join is not executed and init is not called on every box but only one.


Rombur:
My bad apparently the code is useless in serial but not for openmp

Issue #280, opened 2017-08-01
Shape Function CrsMatrix Interpolation Operator


sslattery:
This issue tracks the development of a finite element shape function interpolation operator using Tpetra, Kokkos, and the search capability.

- [ ] Hook mesh interface into coarse search interface

- [ ] Use fine search with results from coarse search

- [ ] Allocate and fill Tpetra CrsMatrix with shape function interpolants using Kokkos

- [ ] Hook field interface into operator apply interface

dalg24:
Closing since #345 and #371 made it into master.  Didn't go with Tpetra matrix though.

PR #303, opened 2017-09-15
Add Interpolation

Rombur:
This PR adds supports for matrix-free Interpolation (relates to #280). The code should have been pretty simple given the already existing capabilities (interface, distributed search, and point in cell) but unfortunately it is not because of the different data structures used by these building blocks. Most of this PR consist of glue code that changes the data structures and of code that keeps track of where the values are (MPI communications and change of data structures means that all the data are reordered several times). There are two different constructors. The first ones allows you to get the position of points in the reference frame. The second constructor also allows you to the interpolation. I strongly advise you to read the PR in the following order the first time:
 1. First [constructor](https://github.com/Rombur/DataTransferKit/blob/integration/packages/Discretization/src/DTK_Interpolation_def.hpp#L29)
 2. [findReferencePoints](https://github.com/Rombur/DataTransferKit/blob/integration/packages/Discretization/src/DTK_Interpolation_def.hpp#L449)
 3. [getReferencePoints](https://github.com/Rombur/DataTransferKit/blob/integration/packages/Discretization/src/DTK_Interpolation_def.hpp#L142)
 4. Second [constructor](https://github.com/Rombur/DataTransferKit/blob/integration/packages/Discretization/src/DTK_Interpolation_def.hpp#L71)
 5. [apply](https://github.com/Rombur/DataTransferKit/blob/integration/packages/Discretization/src/DTK_Interpolation_decl.hpp#L84)

This will give you a global idea of how everything works.

For the tests, the first two functions don't do anything interesting but building two different meshes. The first one is made of simple hex while the second one is mixed topology of triangles and quad. The following four functions are the real tests.


packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    Kokkos::View<int *, DeviceType> cell_indices,
+    Kokkos::View<DataTransferKit::Point *, DeviceType> points,
+    Kokkos::View<int *, DeviceType> query_ids,
+    Kokkos::View<unsigned int *, DeviceType> map,
+    Kokkos::View<int *, DeviceType> filtered_cell_indices,
+    Kokkos::View<double **, DeviceType> filtered_points,
+    Kokkos::View<int *, DeviceType> filtered_query_ids )
+{
+    using ExecutionSpace = typename DeviceType::execution_space;
+    unsigned int const size = topo.extent( 0 );
```
dalg24: This is not a very helpful name.  Is that the total number of cells regardless of their topology?

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+            if ( topo( i ) == topo_id )
+            {
+                unsigned int const k = offset( i );
+                filtered_cell_indices( k ) =
+                    bounding_box_to_cell( cell_indices( i ), topo_id );
+                for ( unsigned int j = 0; j < dim; ++j )
+                {
+                    filtered_points( k, j ) = points( i )[j];
+                    filtered_query_ids( k ) = query_ids( i );
+                    map( k ) = i;
```
dalg24: `map` is also not very descriptive.  Is that used to map the index of the cell in the vector of all cells with mixed topology and map it to the index for the cell in a vector of cells with the same topology?

dalg24: also why are the previous two lines inside the loop over the dimensions?

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    // Perform the distributed search
+    Kokkos::View<int *, DeviceType> indices( "indices" );
+    Kokkos::View<int *, DeviceType> offset( "offset" );
+    Kokkos::View<int *, DeviceType> ranks( "ranks" );
+    distributed_tree.query( queries, indices, offset, ranks );
+
+    // Create the distributor and its reverse that will be used to communicate
+    // the results back to the calling processors.
+    auto ranks_host = Kokkos::create_mirror_view( ranks );
+    Kokkos::deep_copy( ranks_host, ranks );
```
dalg24: I am pretty sure I forgot to do that on my branch...

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    DataTransferKit::DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
+        distributed_search_distributor, indices, imported_cell_indices );
+    // Duplicate the points_coord for the communication. Duplicating the points
+    // allows us to use the same distributor.
+    unsigned int const indices_size = indices.extent( 0 );
+    Kokkos::View<DataTransferKit::Point *, DeviceType> exported_points(
+        "exported_points", indices_size );
+    Kokkos::View<int *, DeviceType> exported_query_ids( "exported_query_ids",
+                                                        indices_size );
+    if ( offset.extent( 0 ) > 0 )
```
dalg24: Can it actually be zero?  I would expect it is going to be of size n_queries + 1.  I remember testing (on node) for search on empty trees but not searching for nothing.  There is definitely no test in the distributed case.

packages/Discretization/src/DTK_Interpolation_decl.hpp
```diff
@@ -0,0 +1,765 @@
...
+        Kokkos::View<DataTransferKit::Point *, DeviceType> &ref_points,
+        Kokkos::View<bool *, DeviceType> &pt_in_cell );
+
+    /**
+     * Convert the 1D Kokkos View cells and coordinates to array of 3D Kokkos
+     * View more suitable for Intrepid2 and the bounding boxes of each cell.
+     *
+     * @note This function should be <b>private</b> but lambda functions can
+     * only be
+     * called from a public function in CUDA.
```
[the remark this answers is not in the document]
Rombur: Yeah that's why we need clang-format 5.0

packages/Discretization/src/DTK_Interpolation_decl.hpp
```diff
@@ -0,0 +1,765 @@
...
+    Kokkos::View<int *, DeviceType> _query_id;
+    /**
+     * Topology associated to each cell.
+     */
+    Kokkos::View<DTK_CellTopology *, DeviceType> _cell_topologies_view;
+    /**
+     * Array of Views of coordinates of the nodes of the cells. Each View in the
+     * array is associated to one topology.
+     */
+    std::array<Kokkos::View<double ***, DeviceType>, DTK_N_TOPO> _block_cells;
```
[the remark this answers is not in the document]
Rombur: Some topologies won't have any cells associated to them so I don't want to have to allocate memory for them.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    _cell_topologies[6] =
+        shards::getCellTopologyData<shards::Tetrahedron<11>>();
+    _cell_topologies[7] = shards::getCellTopologyData<shards::Hexahedron<8>>();
+    _cell_topologies[8] = shards::getCellTopologyData<shards::Hexahedron<20>>();
+    _cell_topologies[9] = shards::getCellTopologyData<shards::Hexahedron<27>>();
+    _cell_topologies[10] = shards::getCellTopologyData<shards::Pyramid<5>>();
+    _cell_topologies[11] = shards::getCellTopologyData<shards::Pyramid<13>>();
+    _cell_topologies[12] = shards::getCellTopologyData<shards::Wedge<6>>();
+    _cell_topologies[13] = shards::getCellTopologyData<shards::Wedge<15>>();
+    _cell_topologies[14] = shards::getCellTopologyData<shards::Wedge<18>>();
```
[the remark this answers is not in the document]
Rombur: Basically I need a map between the enum we have defined and the shards objects. I didn't know where to put it so it is here temporarily.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    for ( unsigned int topo_id = 0; topo_id < DTK_N_TOPO; ++topo_id )
+        n_exported_points += _reference_points[topo_id].extent( 0 );
+
+    Kokkos::View<Point *, DeviceType> exported_ref_points(
+        "exported_ref_points", n_exported_points );
+    Kokkos::View<Point *, DeviceType> exported_phys_points(
+        "exported_phys_points", n_exported_points );
+    Kokkos::View<bool *, DeviceType> exported_pt_in_cell( "exported_pt_in_cell",
+                                                          n_exported_points );
+    unsigned int dim = _dim;
```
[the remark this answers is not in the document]
Rombur: Yes exactly.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+                if ( cell_topo_view( i ) == topo_id )
+                {
+                    DataTransferKit::Box bounding_box;
+                    // If dim == 2, we need to set bounding_box[4] and
+                    // bounding_box[5].
+                    if ( dim == 2 )
+                    {
+                        bounding_box[4] = 0;
+                        bounding_box[5] = 1;
+                    }
```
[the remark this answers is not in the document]
Rombur: No, it doesn't work because in the distributed search, when we compute the Morton code, we will divide (I think) by bounding_box[5]-bounding_box[4].

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    using ExecutionSpace = typename DeviceType::execution_space;
+    Kokkos::View<DataTransferKit::Details::Within *, DeviceType> queries(
+        "queries", n_points );
+    Kokkos::parallel_for( "register_queries",
+                          Kokkos::RangePolicy<ExecutionSpace>( 0, n_points ),
+                          KOKKOS_LAMBDA( int i ) {
+                              queries( i ) = DataTransferKit::Details::within(
+                                  {{points_coord( i, 0 ), points_coord( i, 1 ),
+                                    points_coord( i, 2 )}},
+                                  1e-14 );
```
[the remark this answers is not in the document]
Rombur: Yes, there is a FIXME here. The problem with "overlap" is that it doesn't support point right now.

packages/Discretization/src/DTK_Interpolation_def.hpp
Rombur: It can be zero. In one of the test, there is a processor that does not own any points and does not do any query.

packages/Discretization/src/DTK_Interpolation_def.hpp
Rombur: It's the size of n_imports -> total number of cells where we will do the the point in cell search.

packages/Discretization/src/DTK_Interpolation_def.hpp
Rombur: Yes that is what `map` does. The for loop is too big, you are right.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    Kokkos::fence();
+    Kokkos::parallel_scan(
+        REGION_NAME( "compute_node_offset" ),
+        Kokkos::RangePolicy<ExecutionSpace>( 0, n_cells ),
+        KOKKOS_LAMBDA( int i, int &update, bool final_pass ) {
+            if ( final_pass )
+                node_offset( i ) = update;
+            update += nodes_per_cell( i );
+        } );
+    Kokkos::fence();
```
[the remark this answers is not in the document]
Rombur: I need this function alone. several times later on

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+            }
+        } );
+    Kokkos::fence();
+}
+
+template <typename DeviceType>
+void Interpolation<DeviceType>::computeTopologies()
+{
+    // This needs to be done on the host because we will use _n_cells_per_topo
+    // to allocate Kokkos::View and this needs to be done on the host
```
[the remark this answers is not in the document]
Rombur: Yeah, we could. The reason why `_n_cells_per_topo` is not a Kokkos View is because it is too small to work efficiently on a GPU. However, since we need to copy `_cell_topologies_view` on the cpu, we probably don't gain anything.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    DTK_CellTopology dtk_cell_topo;
+    for ( unsigned int i = 0; i < n_cells; ++i )
+    {
+        dtk_cell_topo = cell_topologies_host( i );
+        _n_cells_per_topo[dtk_cell_topo] += 1;
+    }
+
+    // We do not support meshes that contains both 2D and 3D cells. All the
+    // cells are either 2D or 3D
+    _dim = _cell_topologies[dtk_cell_topo].getDimension();
```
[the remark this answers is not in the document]
Rombur: That's what the comment meant (we assume all the cells are 2D XOR 3D). It needs a test in debug mode.

packages/Discretization/src/DTK_Interpolation_def.hpp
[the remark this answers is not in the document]
dalg24: Nah I think we planned for this.  See [here](https://github.com/ORNL-CEES/DataTransferKit/blob/master/packages/Search/src/details/DTK_DetailsTreeConstruction_def.hpp#L53)

packages/Discretization/src/DTK_Interpolation_def.hpp
dalg24: I need to look more into this.

packages/Discretization/src/DTK_Interpolation_def.hpp
```diff
@@ -0,0 +1,669 @@
...
+    Kokkos::View<double **, DeviceType> nodes_coordinates,
+    Kokkos::View<double **, DeviceType> points_coordinates,
+    Kokkos::View<LocalOrdinal *, DeviceType> object_dof_ids )
+    : Interpolation<DeviceType>( comm, cell_topologies_view, cells,
+                                 nodes_coordinates, points_coordinates )
+{
+    _apply_allowed = true;
+
+    // This function does exactly the same thing as convertMesh but for
+    // object_dof_ids
```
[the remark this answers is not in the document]
Rombur: After talking with @sslattery, we have come to the conclusion that we should get ride of the `shards` objects. However, I would like to do in another PR because it means changes in `PointInCell` and the PR is already large enough so I would prefer not change more files.

packages/Discretization/src/DTK_Interpolation_def.hpp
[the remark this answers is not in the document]
sslattery: I think these will be going away if we get rid of shards

packages/Discretization/src/DTK_Interpolation_def.hpp
dalg24: Per #305 you can get rid of it.

packages/Discretization/test/tstInterpolation.cpp
```diff
@@ -0,0 +1,656 @@
...
+    // Fifth cell
+    cells_host( n++ ) = 5;
+    cells_host( n++ ) = 8;
+    cells_host( n++ ) = 4;
+    // Sixth cell
+    cells_host( n++ ) = 5;
+    cells_host( n++ ) = 6;
+    cells_host( n++ ) = 9;
+    cells_host( n++ ) = 8;
+    Kokkos::deep_copy( cells, cells_host );
```
dalg24: You are missing a great opportunity for ASCII artwork!
```C++
// 7-----8-----9
// |   /   \   |
// 3---4---5---6
// |   \   /   |
// 0-----1-----2
```

packages/Discretization/test/tstInterpolation.cpp
```diff
@@ -0,0 +1,656 @@
...
+    Teuchos::RCP<const Teuchos::Comm<int>> comm =
+        Teuchos::DefaultComm<int>::getComm();
+    int const comm_rank = comm->getRank();
+    unsigned int constexpr dim = 3;
+    Kokkos::View<DTK_CellTopology *, DeviceType> cell_topologies_view;
+    Kokkos::View<unsigned int *, DeviceType> cells;
+    Kokkos::View<double **, DeviceType> coordinates;
+    Kokkos::View<double * [3], DeviceType> points_coord;
+    std::tie( cell_topologies_view, cells, coordinates, points_coord ) =
+        buildHexMesh<DeviceType>( comm );
```
dalg24: Nice!
Let's require C++17 and write
```C++
auto [ cell_topologies_view, cells, coordinates, points_coord ] =
    buildHexMesh<DeviceType>( comm );
```

packages/Discretization/test/tstInterpolation.cpp
dalg24: Nevermind you did it in the unit test :D

packages/Discretization/test/tstInterpolation.cpp
```diff
@@ -0,0 +1,656 @@
...
+        }
+    }
+}
+
+TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( Interpolation, two_topo_two_dim, DeviceType )
+{
+    // Test a mesh of made of Quadrilateral<4> and Triangle<3>
+    // __________
+    // |___/\___|
+    // |___\/___|
```
dalg24: Come on you can do better!

Rombur: You have too high expectations. Not everybody is the Michael Angelo of ascii art like you!

packages/Discretization/test/tstInterpolation.cpp
```diff
@@ -0,0 +1,656 @@
...
+    {
+        TEST_EQUALITY( reference_points.extent( 0 ), 0 );
+    }
+
+    // Reference solution
+    typedef std::array<double, dim> PtCoord;
+    std::map<PtCoord, std::set<PtCoord>> ref_sol;
+    // First point
+    PtCoord pt_1 = {{0.5, 0.5, 0.5}};
+    std::vector<PtCoord> ref_frame_1 = {{{0., 0., 0.}}};
```
dalg24: Stupid Clang warnings...

dalg24: I mean that we cannot write `std::vector<PtCoord> ref_frame_1 = {{0., 0., 0.}};`

packages/Discretization/test/tstInterpolation.cpp
```diff
@@ -0,0 +1,656 @@
...
+                                 phys_points_host( i )[2]}};
+
+            PtCoord ref_pt = {{reference_points_host( i )[0],
+                               reference_points_host( i )[1],
+                               reference_points_host( i )[2]}};
+
+            TEST_EQUALITY( ref_sol[pt_coord].count( ref_pt ), 1 );
+            ref_sol[pt_coord].erase( ref_pt );
+        }
+    }
```
dalg24: You don't want to assert that you erased everything here?

Rombur:
OK done

Rombur:
I have rebased the PR onto the current master

PR #304, opened 2017-09-21
[WIP] Discretization API

sslattery:
This PR covers an updated discretization/field interface for accessing application field data. This includes support for all intrepid shape functions (including HGRAD, HDIV, and HCURL of all orders) as well as user-provided local search and interpolation.

### DIFF
--- a/packages/Interface/src/DTK_CellList.hpp
+++ b/packages/Interface/src/DTK_CellList.hpp
@@ -49,17 +49,15 @@ class CellList
 
     //! Connectivity list of cells. The view rank is rank-1.
     //!
-    //! It represents a lists of cells with different topologies ordered in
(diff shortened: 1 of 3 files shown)


sslattery:
Note the current state will not build but wanted to discuss the API changes @Rombur and I made before they are propagated. Please look at the [interface](https://github.com/sslattery/DataTransferKit/blob/505fe3da37afaffe24d871354fdce37b9dae4a86/packages/Interface/src/DTK_UserDataInterface.hpp) and new [enumerations](https://github.com/sslattery/DataTransferKit/blob/505fe3da37afaffe24d871354fdce37b9dae4a86/packages/Interface/src/DTK_Enumerations.h).

PR #306, opened 2017-09-26
Specify CUDA version when working with containers

dalg24:
Take, respectively, `CUDA_VERSION` and `GPU_ARCH` env variables as default for the `--cuda-version` and `--gpu-arch` arguments in docker/nvidia/setup_nvidia_docker_compose.py

If the args are not passed and the env variables are not here, the script interrogates the nvidia-docker-plugin.

This would let us define these env variables in the Jenkins configuration.

NOTE: Currently we set `CUDA_VERSION=8.0` and `GPU_ARCH=sm_35` for the slave machine (fetnat01)

### DIFF
--- a/docker/jenkins/Makefile
+++ b/docker/jenkins/Makefile
@@ -2,13 +2,22 @@ BASE_FILE=docker-compose.yml
 OVERRIDE_FILE=docker-compose.override.yml
 SERVICE=ci
 SETUP_SCRIPT=../nvidia/setup_nvidia_docker_compose.py
+ENV_VARIABLES=CUDA_VERSION GPU_ARCH
 
 all: setup_nvidia_docker
 
 setup_nvidia_docker: ${OVERRIDE_FILE}
 
-${OVERRIDE_FILE}: ${BASE_FILE} ${SETUP_SCRIPT} Makefile
+${OVERRIDE_FILE}: ${BASE_FILE} ${SETUP_SCRIPT} Makefile ${ENV_VARIABLES}
 	${SETUP_SCRIPT} --extended-file ${BASE_FILE} --extend-services ${SERVICE} --override-file ${OVERRIDE_FILE}
 
 clean:
-	rm -vf *_generated ${OVERRIDE_FILE}
+	rm -vf *_generated ${OVERRIDE_FILE} ${ENV_VARIABLES}
+
+.PHONY: always-rebuild
+
+.SILENT: ${ENV_VARIABLES}
+${ENV_VARIABLES}: always-rebuild
+	echo $@=$$$@ > $@.tmp
+	diff $@ $@.tmp &>/dev/null || cp $@.tmp $@
+	rm -f $@.tmp
--- a/docker/nvidia/setup_nvidia_docker_compose.py
+++ b/docker/nvidia/setup_nvidia_docker_compose.py
@@ -56,6 +56,7 @@ if args.cuda_pkg_version is not None:
 else:
     # from NVIDIA public hub repository https://hub.docker.com/r/nvidia/cuda/
     cuda_pkg_versions = {
+        '9.0': '9-0=9.0.176-1',
         '8.0': '8-0=8.0.61-1',
         '7.5': '7-5=7.5-18',
         '7.0': '7-0=7.0-28',
--- a/docker/nvidia/setup_nvidia_docker_compose.py
+++ b/docker/nvidia/setup_nvidia_docker_compose.py
@@ -14,7 +14,7 @@ with open(dockerfile_template, 'r') as fin:
 parser = argparse.ArgumentParser(description='Enable GPU-aware dev container.')
 parser.add_argument('--nvidia-docker-plugin-url', dest='nvidia_docker_plugin_url', type=str, default='http://localhost:3476',
                     help='nvidia-docker-plugin REST API URL (default: http://localhost:3476)')
-parser.add_argument('--gpu-arch', dest='gpu_arch', type=str,
+parser.add_argument('--gpu-arch', dest='gpu_arch', type=str, default=os.getenv('GPU_ARCH'),
                     help='specify what GPU architecture (default: sm_30)')
 parser.add_argument('--cuda-version', dest='cuda_version', type=str, default=os.getenv('CUDA_VERSION'),
                     help='specify the CUDA version')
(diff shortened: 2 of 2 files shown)


Rombur:
@aprokop If it is not set,  we use the newest version of CUDA that the driver support. The current change is to allow to use CUDA 8 when your driver support CUDA 9.

dalg24:
> Add a check for whether such environment variable exist? If not, do not set the default?

If it does not exist it will be set to `None`.

dalg24:
And yes, as Bruno said, in that case we interrogate nvidia-docker-plugin.

dalg24:
Bugfix: calling `make setup_nvidia_docker` on Jenkins would not capture whether the environment in which the Python script was previously called and not update the extension to the configuration (docker-compose.override.yml file)

dalg24:
This is ready for review/merge.

dalg24:
Also, it seems that Intrepid2 is giving some CUDA 9.0 error (see build [513](https://cloud.cees.ornl.gov/jenkins-ci/job/DataTransferKit-experimental/513/compiler=GCC54,slave=fetnat01/console)) so I am setting CUDA_VERSION back to 8.0.  I will open a separate issue for it.

docker/jenkins/Makefile
```diff
@@ -2,13 +2,22 @@ BASE_FILE=docker-compose.yml
 OVERRIDE_FILE=docker-compose.override.yml
 SERVICE=ci
 SETUP_SCRIPT=../nvidia/setup_nvidia_docker_compose.py
+ENV_VARIABLES=CUDA_VERSION GPU_ARCH
```
sslattery: OK so looking at this now we can specify the cuda version and architecture we are using on Jenkins

docker/jenkins/Makefile
```diff
@@ -2,13 +2,22 @@ BASE_FILE=docker-compose.yml
...
 
 all: setup_nvidia_docker
 
 setup_nvidia_docker: ${OVERRIDE_FILE}
 
-${OVERRIDE_FILE}: ${BASE_FILE} ${SETUP_SCRIPT} Makefile
+${OVERRIDE_FILE}: ${BASE_FILE} ${SETUP_SCRIPT} Makefile ${ENV_VARIABLES}
```
sslattery: and those new variables get passed me make is called

docker/jenkins/Makefile
```diff
@@ -2,13 +2,22 @@ BASE_FILE=docker-compose.yml
...
 	${SETUP_SCRIPT} --extended-file ${BASE_FILE} --extend-services ${SERVICE} --override-file ${OVERRIDE_FILE}
 
 clean:
-	rm -vf *_generated ${OVERRIDE_FILE}
+	rm -vf *_generated ${OVERRIDE_FILE} ${ENV_VARIABLES}
```
sslattery: So this will reset the values of the new environment variables?

docker/jenkins/Makefile
```diff
@@ -2,13 +2,22 @@ BASE_FILE=docker-compose.yml
...
+
+.PHONY: always-rebuild
+
+.SILENT: ${ENV_VARIABLES}
+${ENV_VARIABLES}: always-rebuild
```
sslattery: what does this do? just curious

docker/nvidia/setup_nvidia_docker_compose.py
```diff
@@ -14,9 +14,9 @@
 parser = argparse.ArgumentParser(description='Enable GPU-aware dev container.')
 parser.add_argument('--nvidia-docker-plugin-url', dest='nvidia_docker_plugin_url', type=str, default='http://localhost:3476',
                     help='nvidia-docker-plugin REST API URL (default: http://localhost:3476)')
-parser.add_argument('--gpu-arch', dest='gpu_arch', type=str,
+parser.add_argument('--gpu-arch', dest='gpu_arch', type=str, default=os.getenv('GPU_ARCH'),
```
sslattery: nice

docker/nvidia/setup_nvidia_docker_compose.py
```diff
@@ -56,6 +56,7 @@ def get(self, endpoint):
 else:
     # from NVIDIA public hub repository https://hub.docker.com/r/nvidia/cuda/
     cuda_pkg_versions = {
+        '9.0': '9-0=9.0.176-1',
```
sslattery: so if someone passes `--cuda-version=9.0` then this is the version that will get downloaded from the NVIDIA docker hub?

docker/jenkins/Makefile
dalg24: Yes.  This would enable having both a CUDA 8.0 and CUDA 9.0 build on fetnat01.

docker/jenkins/Makefile
dalg24: Not exactly.  We write stamp files that save the values of these env variables at the time the setup script was last called and add a dependency on these files.

docker/jenkins/Makefile
dalg24: No this is just a convenience rule to remove all the files that are generated.

docker/jenkins/Makefile
dalg24: Every time we call `make setup_nvidia_docker` this rule will be executed and check whether the environment variables we care about changed.

docker/jenkins/Makefile
sslattery: ok - because the variables are put into a file

docker/nvidia/setup_nvidia_docker_compose.py
sslattery: ok - so our images already have these versions. This just tells us which one to pick?

docker/jenkins/Makefile
dalg24: Correct.  And you can inspect the values in the [workspace for that build on Jenkins](https://cloud.cees.ornl.gov/jenkins-ci/job/DataTransferKit-experimental/compiler=GCC54,slave=fetnat01/ws/docker/jenkins/)

docker/nvidia/setup_nvidia_docker_compose.py
sslattery: ok makes sense

sslattery:
@dalg24 I will merge if you are ready

docker/nvidia/setup_nvidia_docker_compose.py
dalg24: No they don't.  We augment our base image and commit into a new image.

dalg24:
Go ahead.

PR #310, opened 2017-10-06
Update Open MPI in base image to be able to run mpiexec from outside the container

dalg24:
I ended up separating this out from #307
This is not ready for review yet.

I tagged a new 17.10.1 image on Docker Hub.  It has Open MPI version 2.1 instead of the 3.0 @Rombur had in his PR.  I want us to test it out with Singularity before we merge.

- [x] experiment with Singularity
- [x] add documentation

### DIFF
--- a/docker/Dockerfile_stack
+++ b/docker/Dockerfile_stack
@@ -1,4 +1,4 @@
-from ubuntu:16.04
+FROM ubuntu:16.04
 
 ARG NPROC=4
 
--- a/docker/Dockerfile_stack
+++ b/docker/Dockerfile_stack
@@ -5,7 +5,6 @@ ARG NPROC=4
 RUN apt-get update && apt-get install -y \
         build-essential \
         gfortran \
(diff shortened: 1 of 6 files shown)


dalg24:
@Rombur I installed 2.1.2 because Singularity claims to support the Open MPI 2.1 series and 2.1.2 was the latest release.  I realized afterwards Condo only had 2.1.1 but my understanding is they will install the newer version.

dalg24:
I tagged the image `17.10.1` as `latest` on DockerHub and dropped the commit that was altering the tag in `docker/jenkins/docker-compose.yml`. This is ready for merge.

@Rombur Should we remove the image tagged `17.09.0` that lives on DockerHub since it got replaced?
```
17.07.1 ---- + ---- 17.09.0
              \
                ---- 17.10.0 ---- 17.10.1
```

Rombur:
@dalg24 yeah I think it makes sense to drop it

PR #314, opened 2017-10-18
Update README for summitdev

Rombur:
mpirun is not accessible on summitdev anymore. It has been replace by jsrun
which does not have an option to enable gpu-aware mpi.


Rombur:
I would prefer not to do it. `jsrun` is still in beta it's better to just go on olcf website to see how to do it. We can add it once it's stable.

PR #321, opened 2017-11-01
Less code more tests for search package

dalg24:


packages/Search/test/tstLinearBVH.cpp
```diff
@@ -41,134 +41,289 @@ void check_results( DataTransferKit::BVH<DeviceType> &bvh,
...
-    void operator()( int const i ) const
-    {
-        if ( i == 0 )
-            _boxes[0] = {{0, 0, 0, 0, 0, 0}};
-        else
-            _boxes[1] = {{1, 1, 1, 1, 1, 1}};
-    }
+template <typename DeviceType>
+DataTransferKit::BVH<DeviceType>
+make_bvh( std::vector<DataTransferKit::Box> const &b )
```
Rombur: Do we want to move all these move all these `make_` functions outside of test?

packages/Search/test/tstLinearBVH.cpp
```diff
@@ -41,134 +41,289 @@ void check_results( DataTransferKit::BVH<DeviceType> &bvh,
...
-    Kokkos::deep_copy( boxes, boxes_host );
-    DataTransferKit::BVH<DeviceType> bvh( boxes );
-    auto bounds = bvh.bounds();
-    TEST_EQUALITY( bounds[0], 1.0 );
-    TEST_EQUALITY( bounds[1], 2.0 );
-    TEST_EQUALITY( bounds[2], 3.0 );
-    TEST_EQUALITY( bounds[3], 4.0 );
-    TEST_EQUALITY( bounds[4], 5.0 );
-    TEST_EQUALITY( bounds[5], 6.0 );
+    // auto bvh = make_bvh<DeviceType>( {
```
Rombur: Why is this test commented?

packages/Search/test/tstLinearBVH.cpp
```diff
@@ -41,134 +41,289 @@ void check_results( DataTransferKit::BVH<DeviceType> &bvh,
...
+                   } ),
+                   {-1, -1, -1}, {0, 1, 3}, success, out );
+
+    double const infty = std::numeric_limits<double>::max();
+    check_results( empty_bvh,
+                   make_nearest_queries<DeviceType>( {
+                       {{{0., 0., 0.}}, 1},
+                       {{{1., 1., 1.}}, 2},
+                   } ),
+                   {-1, -1, -1}, {0, 1, 3}, {infty, infty, infty}, success,
```
Rombur: Should the distance be infty or negative? It seems easier to check that it is negative.

packages/Search/test/tstLinearBVH.cpp
dalg24: Move them out of this test into some header for use in testing for the Search package? Maybe (only if it is actually needed elsewhere).
Make them more broadly accessible? No, these are cheap convenience helpers, they would kill performance and we certainly don't want the user to believe that's the proper way to build them.

packages/Search/test/tstLinearBVH.cpp
dalg24: Forgot to erase it.

packages/Search/test/tstLinearBVH.cpp
dalg24: My personal preference would be to change the behavior and return `{}, {0}, {}`. I want us to have that discussion and I will change the behavior in a separate PR.

packages/Search/test/tstLinearBVH.cpp
dalg24: @Rombur I realize now I must have forgotten to cherry pick one of the commits.  Will update later tonight...

Rombur:
The functions needs to be renamed to satisfy our style guide. I will do it.

dalg24:
Thanks you Bruno

PR #322, opened 2017-11-03
Filter out invalid indices in nearest neighbor search results

dalg24:


packages/Search/src/DTK_LinearBVH_decl.hpp
```diff
@@ -102,12 +102,14 @@ void queryDispatch(
...
     Kokkos::realloc( indices, n_results );
-    fill( indices, -1 );
+    int const invalid_index = -1;
+    Kokkos::deep_copy( indices, invalid_index );
     if ( distances_ptr )
     {
         Kokkos::View<double *, DeviceType> &distances = *distances_ptr;
         Kokkos::realloc( distances, n_results );
-        fill( distances, Kokkos::ArithTraits<double>::max() );
+        double const invalid_distance = -Kokkos::ArithTraits<double>::max();
```
dalg24: @Rombur Note the minus sign here.  It will be filtered out anyway but I changed it based on our discussion on #321

packages/Search/test/tstLinearBVH.cpp
```diff
@@ -190,26 +190,19 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( LinearBVH, empty_tree, DeviceType )
...
 
-    double const infty = std::numeric_limits<double>::max();
     checkResults( empty_bvh,
                   makeNearestQueries<DeviceType>( {
                       {{{0., 0., 0.}}, 1},
                       {{{1., 1., 1.}}, 2},
                   } ),
-                  {-1, -1, -1}, {0, 1, 3}, {infty, infty, infty}, success,
-                  out );
+                  {}, {0, 0, 0}, {}, success, out );
```
dalg24: This change summarizes well the changes introduced in that PR.

dalg24: BEFORE: would return invalid indices/distances when unable to find neighbors.
AFTER: only return what was actually found.

packages/Search/src/details/DTK_DetailsUtils.hpp
```diff
@@ -18,62 +18,92 @@
 namespace DataTransferKit
 {
 
-/** \brief Assigns the given value to all elements in the view.
- */
 template <typename T, typename DeviceType>
-void fill( Kokkos::View<T *, DeviceType> out, T const &value )
+class ExclusiveScanFunctor
```
Rombur: Add a comment explaining why you are using a functor instead of a lambda function

packages/Search/src/details/DTK_DetailsUtils.hpp
```diff
@@ -18,62 +18,92 @@
...
- *  \pre \c in and \c out must have the same size.
+ *  \pre \c src and \c dst must be of rank 1 and have the same size.
  */
-template <typename T, typename DeviceType>
-void exclusivePrefixSum(
-    Kokkos::View<T *, DeviceType> in,
-    Kokkos::View<T *, DeviceType> out = Kokkos::View<T *, DeviceType>() )
+template <typename ST, typename... SP, typename DT, typename... DP>
+void exclusivePrefixSum( Kokkos::View<ST, SP...> const &src,
+                         Kokkos::View<DT, DP...> const &dst )
```
Rombur: `dst` is also a `cons ref`? Looks like the function has no output.

packages/Search/src/details/DTK_DetailsUtils.hpp
```diff
@@ -18,62 +18,92 @@
...
-            update += in_i;
-        } );
+        "exclusive_scan", Kokkos::RangePolicy<ExecutionSpace>( 0, n ),
+        ExclusiveScanFunctor<ValueType, ExecutionSpace>( src, dst ) );
     Kokkos::fence();
 }
 
-/** \brief Get a copy of the last element on the host.
+template <typename T, typename... P>
+void exclusivePrefixSum( Kokkos::View<T, P...> const &v )
```
Rombur: Not a fan of the `const ref`. It looks like you don't have any output.

packages/Search/src/details/DTK_DetailsUtils.hpp
```diff
@@ -18,62 +18,92 @@
...
+                     unsigned( Kokkos::ViewTraits<ST, SP...>::rank ) ) &&
+                       ( unsigned( Kokkos::ViewTraits<DT, DP...>::rank ) ==
+                         unsigned( 1 ) ),
+                   "exclusivePrefixSum requires Views of rank 1" );
+
+    using ExecutionSpace =
+        typename Kokkos::ViewTraits<DT, DP...>::execution_space;
+    using ValueType = typename Kokkos::ViewTraits<DT, DP...>::value_type;
+
+    auto const n = src.span();
```
Rombur: Why aren't you using `extent(0)`?

packages/Search/src/details/DTK_DetailsUtils.hpp
dalg24: Yes, view is const but values are not.  `Kokkos::deep_copy()` also takes both views as const ref.

packages/Search/src/details/DTK_DetailsUtils.hpp
dalg24: I never liked `extent( 0 )`.  Here we ensure that the rank is one which makes `span() == extent(0)`

packages/Search/src/details/DTK_DetailsUtils.hpp
Rombur: I know, still ugly to read. It looks like there is no output.

packages/Search/src/details/DTK_DetailsUtils.hpp
dalg24: What would you prefer?  By value?

Rombur: Just remove the `const` so it's obvious that the view is changed.

packages/Search/src/details/DTK_DetailsUtils.hpp
dalg24: I added a parameter description with `[in]` and `[out]` specifiers.

PR #327, opened 2017-11-11
Improve distributed search tree

dalg24:
Summary of what this PR does:
* Let `BVH` default-construct to an empty tree and remove `shared_ptr` in `DistributedSearchTree`
* Compute `_size` in the constructor of `DistributedSearchTree` (give up on lazy evaluation for now since it was forcing us to write more code than it is worth to ensure thread-safety)
* Cleanup unit testing for `DistributedSearchTree` and add more tests
* Introduced common header with helper functions to create trees and queries and check results


dalg24:
Will rebase onto #326

dalg24:
This is ready for review.  Let me know how much you want me to rework history on this PR.
I am tempted to keep the two commits with the copy/paste marked with `[do not merge]` (I would remove the `[...]`).  Let me know what you think.

packages/Search/src/DTK_LinearBVH_decl.hpp
```diff
@@ -33,6 +33,7 @@ template <typename DeviceType>
 class BVH
 {
   public:
+    BVH() = default;
```
Rombur: Add a comment to say that this constructor should not be used since nothing useful can be done with this constructor.

packages/Search/test/Search_UnitTestHelpers.hpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    // NOTE: `points` is not a very descriptive name here. It stores both the
+    // actual point and the number k of neighbors to query for.
+    int const n = points.size();
+    Kokkos::View<DataTransferKit::Details::Nearest *, DeviceType> queries(
+        "nearest_queries", n );
+    auto queries_host = Kokkos::create_mirror_view( queries );
+    for ( int i = 0; i < n; ++i )
+        queries_host( i ) = DataTransferKit::Details::nearest(
+            points[i].first, points[i].second );
```
Rombur: What happens if `points[i].second` is -1?

packages/Search/test/Search_UnitTestHelpers.hpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    // NOTE: `points` is not a very descriptive name here. It stores both the
+    // actual point and the radius for the search around that point.
+    int const n = points.size();
+    Kokkos::View<DataTransferKit::Details::Within *, DeviceType> queries(
+        "within_queries", n );
+    auto queries_host = Kokkos::create_mirror_view( queries );
+    for ( int i = 0; i < n; ++i )
+        queries_host( i ) = DataTransferKit::Details::within(
+            points[i].first, points[i].second );
```
Rombur: What happens if  `points[i].second < 0`?

packages/Search/test/tstDistributedSearchTree.cpp
```diff
@@ -146,109 +145,146 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DistributedSearchTree, hello_world,
...
+                      makeOverlapQueries<DeviceType>( {
+                          {},
+                          {},
+                      } ),
+                      {}, {0, 0, 0}, {}, success, out );
+    else
+        checkResults( empty_tree, makeOverlapQueries<DeviceType>( {} ), {}, {0},
+                      {}, success, out );
+
+    // All ranks but rank 0 have a single query with a single query with a
```
Rombur: Typo `a single query with a single query with a`

packages/Search/test/Search_UnitTestHelpers.hpp
dalg24: The search result will have `offset(i) == offset(i+1)`, i.e. it will tell you it did not find anything.
The construction of the nearest predicate will not complain since it only stores `k` and the query will not traverse the tree and return (See [here](https://github.com/ORNL-CEES/DataTransferKit/blob/master/packages/Search/src/details/DTK_DetailsTreeTraversal.hpp#L173-L199)

packages/Search/test/Search_UnitTestHelpers.hpp
dalg24: Same here, the predicate will not complain and will always evaluate all nodes to `false`.  The query on the tree will only process the root node and will exit directly after since the predicate won't be satisfied on the children (see [here](https://github.com/ORNL-CEES/DataTransferKit/blob/master/packages/Search/src/details/DTK_DetailsPredicate.hpp#L86-L91) and [here](https://github.com/ORNL-CEES/DataTransferKit/blob/master/packages/Search/src/details/DTK_DetailsTreeTraversal.hpp#L105-L130)). We may check the predicate on the root before adding it to the stack if you like but I don't think it is a big deal.

packages/Search/test/tstDistributedSearchTree.cpp
dalg24: Will fix.

PR #331, opened 2017-11-16
Compile boost with rpath support

Rombur:
By default, Boost does not support rpath. This is a problem if a boost library
depends on another one (eg. boost.mpi depends on boost.serialization) and boost
lib is not in the LD_LIBRARY_PATH.

### DIFF
--- a/docker/.vimrc
+++ b/docker/.vimrc
@@ -1,3 +1,5 @@
+set encoding=utf-8
+
 set nocompatible              " be iMproved, required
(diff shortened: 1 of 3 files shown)


dalg24:
Done. I also updated the CHANGELOG in `docker/README.md`.

dalg24:
I just checked and the image has been tagged properly and pushed to Docker Hub.

dalg24:
Thank you @Rombur

PR #320, opened 2017-10-31
[WIP] View and DualView Utilities

sslattery:
Summary of PR:

1. Adds tests for using `Kokkos::View` and `Kokkos::DualView` together to understand the effects of mirroring and how to translate between the two view types
2. Adds tests for using `Kokkos::DualView` with `Tpetra::MultiVector` to understand how to manipulate the vector data directly with the views
3. Adds tests for `Tpetra::Map` and `Tpetra::MultiVector` to understand how to move between various combinations of uniquely owned and ghosted decompositions. This resolves how to do a ghosted to ghosted transfer.
4. Defines utilities that allow for creation of `Kokkos::DualView' objects from regular `Kokkos::View` objects and vice versa
5. Defines utilities that decorate ghosted `Kokkos::View` objects to allow for the construction and import/export of unique views.

PR #336, opened 2017-12-01
Mesh Free Problem Generator

sslattery:
Summary of PR:

1. Defines an interface to use for point cloud test problem generation. This includes parallel point coordinates, point ids, and field data.
2. Add an implementation of the generator for exodus files. The generator reads the exodus files, partitions the nodes of the elements in the mesh, and returns the problem.
3. Defines tests for the current mesh free operators using the exodus problem generator.

Notes:

- The problem generator builds both uniquely owned and ghosted sets of points with corresponding unique global ids.
- The unit test for `DataTransferKit::NearestNeighborOperator` only uses the unique partitioning of points as the API only takes uniquely partitioned points. The ghosted problem generator will be used when we write tests for the decorators that transition ghosted points to uniquely owned points.

To Do:

- [x] Add ghosted operator test.
- [x] Fix build dependency so generator tests only build when netcdf is available and other tests always build when tests are on.

### DIFF
--- a/packages/Meshfree/cmake/Dependencies.cmake
+++ b/packages/Meshfree/cmake/Dependencies.cmake
@@ -1,10 +1,12 @@
 TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
   LIB_REQUIRED_PACKAGES
   DataTransferKitUtils
+  DataTransferKitInterface
   DataTransferKitSearch
   Kokkos
   Teuchos
   Tpetra
   TEST_OPTIONAL_TPLS
   BoostOrg
+  Netcdf
   )
--- a/packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
+++ b/packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
@@ -1,0 +1,162 @@
+/****************************************************************************
+ * Copyright (c) 2012-2018 by the DataTransferKit authors                   *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the DataTransferKit library. DataTransferKit is     *
+ * distributed under a BSD 3-clause license. For the licensing terms see    *
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef DTK_EXODUSPROBLEMGENERATOR_HPP
+#define DTK_EXODUSPROBLEMGENERATOR_HPP
+
+#include "DTK_ConfigDefs.hpp"
+#include "DTK_Types.h"
+#include "PointCloudProblemGenerator.hpp"
+
+#include <Kokkos_View.hpp>
+
+#include <Teuchos_Comm.hpp>
+#include <Teuchos_RCP.hpp>
+
+#include <functional>
+#include <string>
+
+namespace DataTransferKit
+{
+//---------------------------------------------------------------------------//
+// Generate point cloud problem by reading exodus files.
+//
+// The generator reads exodus files and extracts the node coordinates and
+// partitions them across the given communicator.
+//
+// Source files are partitioned in x where each rank gets an even subdivision
+// of space in the x dimension:
+//
+// |          | o        |        |                      |
+// |  o     o |       o  |  o  o  |  o o       o         |
+// |    o     |  o     o |      o |        o      o      |
+// |   o   o  |     o    | o      |     o    o  o        |
+// |          |          |    o   |                      |
+// | rank = 0 | rank = 1 | ...... | rank = comm_size - 1 |
+//
+// Target files are partitioned in y where each rank gets an even subdivision
+// of space in the y dimension.
+//
+// ---------------------------------
+//             o o   o     o
+// rank = 0       o     o   o
+//              o   o   o o
+// ---------------------------------
+//               o   o    o   o
+// rank = 1    o  o   o o    o
+//              o    o oo    o o
+// ---------------------------------
+//               oo  o o o o  oo
+// ....       o   o  o  o  o  o
+//              o  oo  o o oo o o
+// ---------------------------------
+//                   o o o oo
+// rank =             o   o o o
+//   comm_size - 1   o   oo  ooooo
+//
+// ---------------------------------
+//
+// In the uniquely owned case, mesh nodes are given one single, unique
+// destination in the partitioning.
+//
+// In the ghosted case, mesh elements are given one unique destination and
+// all nodes belonging to that element are sent to that
+// destination. Therefore, nodes owned by multiple elements with different
+// destinations will be sent to multiple ranks and thus ghosted.
+//
+template <class Scalar, class SourceDevice, class TargetDevice>
+class ExodusProblemGenerator
+    : public PointCloudProblemGenerator<Scalar, SourceDevice, TargetDevice>
+{
+  public:
+    // Constructor.
+    ExodusProblemGenerator( const Teuchos::RCP<const Teuchos::Comm<int>> &comm,
+                            const std::string &source_exodus_file,
+                            const std::string &target_exodus_file );
+
+    // Create a problem where all points are uniquely owned (i.e. no
+    // ghosting). Both source and target fields have one component and are
+    // initialized to zero.
+    void createUniquelyOwnedProblem(
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, SourceDevice>
+            &src_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, SourceDevice> &src_field,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, TargetDevice>
+            &tgt_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, TargetDevice> &tgt_field )
+        override;
+
+    // Create a general problem where points may exist on multiple
+    // processors. Both source and target fields have 1 component and are
+    // initialized to zero.
+    void createGhostedProblem(
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, SourceDevice>
+            &src_coords,
+        Kokkos::View<GlobalOrdinal *, Kokkos::LayoutLeft, SourceDevice>
+            &src_gids,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, SourceDevice> &src_field,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, TargetDevice>
+            &tgt_coords,
+        Kokkos::View<GlobalOrdinal *, Kokkos::LayoutLeft, TargetDevice>
+            &tgt_gids,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, TargetDevice> &tgt_field )
+        override;
+
+  private:
+    // Get host views of node data from file.
+    void getNodeDataFromFile( const std::string &exodus_file,
+                              Kokkos::View<Coordinate **, Kokkos::LayoutLeft,
+                                           Kokkos::Serial> &host_coords,
+                              Kokkos::View<GlobalOrdinal *, Kokkos::LayoutLeft,
+                                           Kokkos::Serial> &host_gids );
+
+    // Partition a point cloud in a given dimension with one-to-one mapping.
+    template <class Device>
+    void partitionUniquelyOwned( const int dim, const std::string &exodus_file,
+                                 Kokkos::View<Coordinate **, Kokkos::LayoutLeft,
+                                              Device> &device_coords );
+
+    // Partition a point cloud in a given dimension with ghosted connectivity
+    // mapping.
+    template <class Device>
+    void partitionGhostedConnectivity(
+        const int dim, const std::string &exodus_file,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, Device> &device_coords,
+        Kokkos::View<GlobalOrdinal *, Kokkos::LayoutLeft, Device>
+            &device_gids );
+
+    // Given a netcdf handle and a dimension name get the length of that
+    // dimension.
+    size_t getNetcdfDimensionLength( const int nc_id,
+                                     const std::string &dim_name );
+
+  private:
+    // Comm
+    Teuchos::RCP<const Teuchos::Comm<int>> _comm;
+
+    // Filenames
+    std::string _src_exodus_file;
+    std::string _tgt_exodus_file;
+};
+
+//---------------------------------------------------------------------------//
+
+} // end namespace DataTransferKit
+
+//---------------------------------------------------------------------------//
+// Template includes
+//---------------------------------------------------------------------------//
+
+#include "ExodusProblemGenerator_def.hpp"
+
+//---------------------------------------------------------------------------//
+
+#endif // end  DTK_EXODUSPROBLEMGENERATOR_HPP
--- a/packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator_def.hpp
+++ b/packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator_def.hpp
@@ -1,0 +1,409 @@
+/****************************************************************************
+ * Copyright (c) 2012-2018 by the DataTransferKit authors                   *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the DataTransferKit library. DataTransferKit is     *
+ * distributed under a BSD 3-clause license. For the licensing terms see    *
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef DTK_EXODUSPROBLEMGENERATOR_DEF_HPP
+#define DTK_EXODUSPROBLEMGENERATOR_DEF_HPP
+
+#include <DTK_DetailsDistributedSearchTreeImpl.hpp>
+#include <DTK_DetailsUtils.hpp>
+
+#include <netcdf.h>
+
+#include <Teuchos_Array.hpp>
+
+#include <Tpetra_Distributor.hpp>
+
+#include <DTK_DBC.hpp>
+
+#include <fstream>
+#include <iostream>
+#include <limits>
+#include <set>
+
+namespace DataTransferKit
+{
+//---------------------------------------------------------------------------//
+template <class Scalar, class SourceDevice, class TargetDevice>
+ExodusProblemGenerator<Scalar, SourceDevice, TargetDevice>::
+    ExodusProblemGenerator( const Teuchos::RCP<const Teuchos::Comm<int>> &comm,
+                            const std::string &source_exodus_file,
+                            const std::string &target_exodus_file )
+    : _comm( comm )
+    , _src_exodus_file( source_exodus_file )
+    , _tgt_exodus_file( target_exodus_file )
+{ /* ... */
+}
+
+//---------------------------------------------------------------------------//
+// Create a problem where all points are uniquely owned (i.e. no ghosting)
+template <class Scalar, class SourceDevice, class TargetDevice>
+void ExodusProblemGenerator<Scalar, SourceDevice, TargetDevice>::
+    createUniquelyOwnedProblem(
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, SourceDevice>
+            &src_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, SourceDevice> &src_field,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, TargetDevice>
+            &tgt_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, TargetDevice> &tgt_field )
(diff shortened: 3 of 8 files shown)


packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    // On each rank get the source coordinates with no partitioning and find
+    // the nearest neighbors to the locally-owned target points.
+    for ( int r = 0; r < comm->getSize(); ++r )
+    {
+        Kokkos::View<DataTransferKit::Coordinate **, Kokkos::LayoutLeft, Device>
+            src_coords;
+
+        // Load the source coordinates.
+        {
```
dalg24: Why do you have a curly brace around that bock?  Are you introducing a new scope on purpose to shadow the `tgt_coords` argument???  I was wondering how the call to `createUniquelyOwnedProblem()` even compiled since `tgt_coords` is a const reference in the declaration of `computeNeighborsBruteForce()`...

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,259 @@
...
+//---------------------------------------------------------------------------//
+// Brute-force calculate the nearest neighbors
+template <class Device>
+void computeNeighborsBruteForce(
+    const Teuchos::RCP<const Teuchos::Comm<int>> &comm,
+    const std::string &src_exodus_file, const std::string &tgt_exodus_file,
+    const Kokkos::View<DataTransferKit::Coordinate **, Kokkos::LayoutLeft,
+                       Device> &tgt_coords,
+    Kokkos::View<DataTransferKit::Coordinate **, Kokkos::LayoutLeft,
+                 Kokkos::Serial> &src_nearest_coords )
```
dalg24: Why do you work from file names rather than from `src_coords` and `tgt_coords`?

packages/Meshfree/cmake/Dependencies.cmake
```diff
@@ -5,4 +5,6 @@ TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
   Kokkos
   Teuchos
   Tpetra
+  TEST_REQUIRED_TPLS
```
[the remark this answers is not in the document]
sslattery: Not quite sure on the CMake logic here. Lets talk about this tomorrow

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator_def.hpp
```diff
@@ -0,0 +1,416 @@
...
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, SourceDevice> &src_field,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, TargetDevice>
+            &tgt_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, TargetDevice> &tgt_field )
+{
+    // Partition the source coordinates in the x direction.
+    partitionUniquelyOwned( 0, _src_exodus_file, src_coords );
+
+    // Partition the target coordinates in the y direction.
+    partitionUniquelyOwned( 1, _tgt_exodus_file, tgt_coords );
```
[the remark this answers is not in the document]
sslattery: done

packages/Meshfree/test/PointCloudProblemGenerator/PointCloudProblemGenerator.hpp
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * \param tgt_coords Coordinates of the target points. Layout:
+     * (point,dim).
+     *
+     * \param tgt_field Field values at the target points. A field may have
+     * multiple components. Will be filled with zeros. Layout: (point,comp).
+     */
+    virtual void createUniquelyOwnedProblem(
+        const int field_dim,
+        Kokkos::View<Coordinate **, Kokkos::LayoutLeft, SourceDevice>
```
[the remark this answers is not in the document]
sslattery: Two reasons. 1) If we wanted to switch to a `Tpetra::MultiVector` interface we would require views of `LayoutLeft` to directly construct vectors from the views. 2) Coordinates are stored in block contiguous format in the exodus files so it was easy to directly load the coordinates from the exodus file into the view.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
sslattery: To make it easy I just wanted to reload the coordinates in a serial partitioning on each rank. This obviously means doing some extra work but it was an easy way to get the coordinates.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
sslattery: yeah that is wrong shadowing is wrong - I have fixed this. The bracket scope is just so the dummy variables (who's names have been changed to reflect this) don't live after the loading of the exodus files.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,259 @@
...
+    using DeviceType = typename Node::device_type;
+    using ExecutionSpace = typename DeviceType::execution_space;
+    using Scalar = double;
+
+    // Get the communicator.
+    auto comm = Teuchos::DefaultComm<int>::getComm();
+
+    // Create a problem generator.
+    std::string src_exodus_file = "fine_sphere.exo";
+    std::string tgt_exodus_file = "coarse_sphere.exo";
```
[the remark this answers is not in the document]
sslattery: done

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+//---------------------------------------------------------------------------//
+
+// Include the test macros.
+#include "DataTransferKitMeshfree_ETIHelperMacros.h"
+
+// Create the test group
+#define UNIT_TEST_GROUP( NODE )                                                \
+    TEUCHOS_UNIT_TEST_TEMPLATE_1_INSTANT( ExodusProblemGenerator, one_to_one,  \
+                                          NODE )
```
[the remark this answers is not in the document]
sslattery: Not yet. `NearestNeighborOperator` only takes uniquely partitioned coordinate sets. When we have decorators written that handle general overlap partitioning we will use the ghosted problem generator.

packages/Meshfree/test/PointCloudProblemGenerator/PointCloudProblemGenerator.hpp
```diff
@@ -55,6 +58,8 @@ class PointCloudProblemGenerator
      * \brief Create a general problem where points may exist on multiple
      * processors. Points have a unique global id.
      *
+     * \param field_dim The number of components in the field.
```
dalg24: This input parameter does not make sense to me.  The user can infer the number of components from `src_field.extent( 0 )`.  How many component is up to the derived class.
As for `ExodusProblemGenerator`, I do understand that you need it there because you are not actually reading the fields from the files and you are just allocating memory and filling with zero.  But in that case you can always pass `field_dim` to the constructor.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -125,7 +126,16 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( ExodusProblemGenerator, one_to_one, Node )
     // Get the communicator.
     auto comm = Teuchos::DefaultComm<int>::getComm();
 
-    // Create a problem generator.
+    // Set an epsilon for the search. The meshes (see below) mesh a sphere of
+    // radius 10. An epsilon of 1.0 is needed to fully capture nearest
```
dalg24: I don't like "is needed".  Maybe "is sufficient in order to".

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -188,11 +198,6 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( ExodusProblemGenerator, one_to_one, Node )
                           tgt_coord_copy );
     Kokkos::fence();
 
-    // Set an epsilon for the search.
-    auto const epsilon_default =
-        DataTransferKit::DistributedSearchTreeImpl<DeviceType>::epsilon;
-    DataTransferKit::DistributedSearchTreeImpl<DeviceType>::epsilon = 1.0;
-
```
dalg24: I actually like it better just before the operator.   Note that it is only used at construction time so you may set it back to the default value right after.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -205,8 +210,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( ExodusProblemGenerator, one_to_one, Node )
                                                          num_src );
     auto src_field_copy = KOKKOS_LAMBDA( const size_t i )
     {
-        for ( size_t n = 0; n < field_dim; ++n )
-            src_field_device( i ) = src_field( i, n );
+        src_field_device( i ) = src_field( i, 0 );
     };
```
dalg24: Also you do not use the lambda anywhere else you don't need to make an object, you can directly declare it in the `parallel_for()`.

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
dalg24: I do not see how it is easier than passing the two point cloud since you have already parsed the files and then possibly use Tpetra import/export capability if you want to import the whole source vector on the current rank.

sslattery: Right - I can refactor to do the import/export

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,434 @@
...
+    auto comm = Teuchos::DefaultComm<int>::getComm();
+
+    // Create a problem generator. Two tet meshes of a sphere centered at
+    // (0,0,0) with a radius of 10 are used. The source mesh is represented by
+    // a fine tet mesh and the target mesh is represented by a coarse tet mesh.
+    std::string src_exodus_file = "fine_sphere.exo";
+    std::string tgt_exodus_file = "coarse_sphere.exo";
+    auto generator = std::make_shared<DataTransferKit::ExodusProblemGenerator<
+        Scalar, DeviceType, DeviceType>>( comm, src_exodus_file,
+                                          tgt_exodus_file );
```
dalg24: Why using a smart pointer here?

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,434 @@
...
+    Kokkos::View<DataTransferKit::Coordinate **, Device> src_coords_device(
+        "src_coords_device", num_src, space_dim );
+    auto src_coord_copy = KOKKOS_LAMBDA( const size_t i )
+    {
+        for ( size_t n = 0; n < space_dim; ++n )
+            src_coords_device( i, n ) = src_coords( i, n );
+    };
+    Kokkos::parallel_for( Kokkos::RangePolicy<ExecutionSpace>( 0, num_src ),
+                          src_coord_copy );
+    Kokkos::fence();
```
dalg24: Why wouldn't you just do `Kokkos::deep_copy( src_coords_device, src_coords )`?

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -0,0 +1,434 @@
...
+    Kokkos::deep_copy( host_tgt_field, tgt_field_device );
+    double data_val;
+    for ( size_t t = 0; t < num_tgt; ++t )
+    {
+        data_val =
+            data_func( nearest_src_coords( t, 0 ), nearest_src_coords( t, 1 ),
+                       nearest_src_coords( t, 2 ), 0 );
+        test_result( t, 0 ) =
+            std::abs( host_tgt_field( t ) - data_val ) < 1.0e-12;
+    }
```
dalg24: Why wouldn't you just do your assertions in here?

packages/Meshfree/test/PointCloudProblemGenerator/PointCloudProblemGenerator.hpp
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * \param field_dim The number of components in the field.
+     *
+     * \param src_coords Coordinates of the source points. Layout:
+     * (point,dim).
+     *
+     * \param src_field Field values at the source points. A field may have
+     * multiple components. A generator implementation may choose to fill this
+     * with meaningful field data or may choose to initialize it to
+     * zero. Layout: (point,comp).
```
dalg24: Documentation got out of sync.  Also, why did you remove the field values all together?  I recall in previous review only asking for `field_dim` parameter to be removed.

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator_def.hpp
```diff
@@ -0,0 +1,397 @@
...
+    // Get the node data.
+    Kokkos::View<Coordinate **, Kokkos::LayoutLeft, Kokkos::Serial>
+        export_coords;
+    Kokkos::View<GlobalOrdinal *, Kokkos::LayoutLeft, Kokkos::Serial>
+        export_gids;
+    getNodeDataFromFile( exodus_file, export_coords, export_gids );
+
+    // Figure out the min and max coordinates in the given dimension.
+    Coordinate dim_max, dim_min;
+    dimensionMinAndMax( dim, export_coords, dim_min, dim_max );
```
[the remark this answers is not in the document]
sslattery: It would be nice to move `DTK_DetailsUtils.hpp` into the `Utils` package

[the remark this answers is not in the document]
sslattery: In fact that makes sense to do within the context of #320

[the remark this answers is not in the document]
dalg24: > It would be nice to move `DTK_DetailsUtils.hpp` into the `Utils` package

Will do it at some point.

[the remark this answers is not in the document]
dalg24: > In fact that makes sense to do within the context of #320

What do you mean?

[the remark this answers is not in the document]
sslattery: I see the code in `DTK_DetailsUtils.hpp` as a nice set of general utilities for `Kokkos::View` objects. Although admittedly this is additional scope over the definition of #320

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
sslattery: I agree but it doesn't compile:
```
[ 98%] Building CXX object DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/tstNearestNeighborExodusGenerator.cpp.o
/scratch/source/trilinos/release/DataTransferKit/packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp(267): error: identifier "out" is undefined
          detected during instantiation of "void ExodusProblemGenerator_one_to_one_UnitTest<Node>::runUnitTestImpl(Teuchos::FancyOStream &, __nv_bool &) const [with Node=Kokkos_Compat_KokkosSerialWrapperNode]" 
(403): here

/scratch/source/trilinos/release/DataTransferKit/packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp(267): error: identifier "success" is undefined
          detected during instantiation of "void ExodusProblemGenerator_one_to_one_UnitTest<Node>::runUnitTestImpl(Teuchos::FancyOStream &, __nv_bool &) const [with Node=Kokkos_Compat_KokkosSerialWrapperNode]" 
(403): here

2 errors detected in the compilation of "/tmp/tmpxft_0000292c_00000000-7_tstNearestNeighborExodusGenerator.cpp1.ii".
DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/build.make:62: recipe for target 'DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/tstNearestNeighborExodusGenerator.cpp.o' failed
make[2]: *** [DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/tstNearestNeighborExodusGenerator.cpp.o] Error 2
CMakeFiles/Makefile2:5514: recipe for target 'DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/all' failed
make[1]: *** [DataTransferKit/packages/Meshfree/test/CMakeFiles/DataTransferKitMeshfree_NearestNeighborExodusGenerator.dir/all] Error 2
Makefile:160: recipe for target 'all' failed
make: *** [all] Error 2
```

Ultimately I think this is due to the fact that this is a templated unit test and the macros just can't figure it out.

dalg24: Add the `Teuchos::FancyOStream &out` and `bool &success` arguments to your function.

packages/Meshfree/test/PointCloudProblemGenerator/PointCloudProblemGenerator.hpp
sslattery: So the big issue with the field being returned by the generator is also providing the expected result on the target points. Do we return the expected target result? 

I would argue the correct result depends on the operator one uses. So, for example, the nearest operator correct result is different from a standard interpolation. 

On the opposite side, one could argue that these generator tests just check "how close" we are to some expected solution. The nearest neighbor test would then check the results relative to some large tolerance.

@dalg24 what do you think?

dalg24: Generator should at the very least allocate memory and returns properly sized view for the target values.  When it is an option it should fill the view with expected data.  In the context of testing, then it will let us assert that the answer is close enough.

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
```diff
@@ -81,24 +81,32 @@ class ExodusProblemGenerator
...
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, SourceDevice> &src_field,
         Kokkos::View<Coordinate **, Kokkos::LayoutLeft, TargetDevice>
-            &tgt_coords ) override;
+            &tgt_coords,
+        Kokkos::View<Scalar **, Kokkos::LayoutLeft, TargetDevice> &tgt_field )
+        override;
 
     // Create a general problem where points may exist on multiple
-    // processors. Both source and target fields are initialized to zero.
+    // processors. Both source and target fields have 1 component and are
```
dalg24: Why just one?  I thought exodus files could include solution files.

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
```diff
@@ -0,0 +1,160 @@
+/****************************************************************************
+ * Copyright (c) 2012-2017 by the DataTransferKit authors                   *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the DataTransferKit library. DataTransferKit is     *
+ * distributed under a BSD 3-clause license. For the licensing terms see    *
+ * the LICENSE file in the top-level directory.                             *
+ ****************************************************************************/
```
dalg24: Please update headers as in #359

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
sslattery: I wasn't aware of solution files.

In this case the basic exodus generator just reads a mesh and allocates a field of a single component as it has no internal solution state. Other implementations of the interface could return fields of multiple components. If we generate some exodus solution files in the future we could refactor this to return fields with dimensions that we get from the solutions.

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator.hpp
sslattery: good catch

sslattery: done

dalg24: You are missing the SPDX identifier

sslattery: ah right forgot that

PR #338, opened 2017-12-06
More unrelated changes to Search package

dalg24:
* Add Sphere geometric object
* Rename `overlaps(box, box)` algorithm to `intersects()`
* Replace `Overlap` and `Within` predicates by `Intersects<Box>` and `Intersects<Sphere>`
* Template `Nearest` predicate on `Geometry`
* Improve `TreeTraversal::nearestQuery()`

### DIFF
--- a/packages/Discretization/test/tstMeshGenerator.cpp
+++ b/packages/Discretization/test/tstMeshGenerator.cpp
@@ -9,6 +9,7 @@
 
 #include "MeshGenerator.hpp"
 #include <Teuchos_UnitTestHarness.hpp>
+#include <fstream>
 
 std::tuple<std::vector<std::vector<Coordinate>>, std::vector<unsigned int>>
 readInputFile( std::string const &filename )
--- a/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
+++ b/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
@@ -11,6 +11,7 @@
 
 #include <DTK_DetailsBox.hpp>
 #include <DTK_DetailsPoint.hpp>
+#include <DTK_DetailsSphere.hpp>
 #include <DTK_KokkosHelpers.hpp>
 
 #include <Kokkos_Macros.hpp>
@@ -34,6 +35,12 @@ bool equals( Box const &l, Box const &r )
 {
     return equals( l.minCorner(), r.minCorner() ) &&
            equals( l.maxCorner(), r.maxCorner() );
+}
+
+KOKKOS_INLINE_FUNCTION
+bool equals( Sphere const &l, Sphere const &r )
+{
+    return equals( l.centroid(), r.centroid() ) && l.radius() == r.radius();
 }
 
 // distance point-point
--- a/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
+++ b/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
@@ -73,6 +73,14 @@ double distance( Point const &point, Box const &box )
     return distance( point, projected_point );
 }
 
+// distance point-sphere
+KOKKOS_INLINE_FUNCTION
+double distance( Point const &point, Sphere const &sphere )
+{
+    return KokkosHelpers::max(
+        distance( point, sphere.centroid() ) - sphere.radius(), 0. );
+}
+
 // expand an axis-aligned bounding box to include a point
 KOKKOS_INLINE_FUNCTION
 void expand( Box &box, Point const &point )
--- a/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
+++ b/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
@@ -87,15 +87,17 @@ void expand( Box &box, Point const &point )
 }
 
 // expand an axis-aligned bounding box to include another box
-KOKKOS_INLINE_FUNCTION
-void expand( Box &box, Box const &other )
+template <typename BOX,
+          typename = std::enable_if<std::is_same<
+              typename std::remove_volatile<BOX>::type, Box>::value>>
+KOKKOS_INLINE_FUNCTION void expand( BOX &box, BOX const &other )
 {
     for ( int d = 0; d < 3; ++d )
     {
-        if ( box.minCorner()[d] > other.minCorner()[d] )
-            box.minCorner()[d] = other.minCorner()[d];
-        if ( box.maxCorner()[d] < other.maxCorner()[d] )
-            box.maxCorner()[d] = other.maxCorner()[d];
+        box.minCorner()[d] =
+            KokkosHelpers::min( box.minCorner()[d], other.minCorner()[d] );
+        box.maxCorner()[d] =
+            KokkosHelpers::max( box.maxCorner()[d], other.maxCorner()[d] );
     }
 }
 
@@ -131,7 +133,7 @@ void centroid( Box const &box, Point &c )
         c[d] = 0.5 * ( box.minCorner()[d] + box.maxCorner()[d] );
 }
 
-// FIXME: use expand()
+// TODO: move this to Details::TreeConstruction<DeviceType>
 template <typename DeviceType>
 class ExpandBoxWithBoxFunctor
 {
@@ -157,25 +159,13 @@ class ExpandBoxWithBoxFunctor
     KOKKOS_INLINE_FUNCTION
     void operator()( int const i, Box &box ) const
     {
-        for ( int d = 0; d < 3; ++d )
-        {
-            box.minCorner()[d] = KokkosHelpers::min(
-                _bounding_boxes[i].minCorner()[d], box.minCorner()[d] );
-            box.maxCorner()[d] = KokkosHelpers::max(
-                _bounding_boxes[i].maxCorner()[d], box.maxCorner()[d] );
-        }
+        expand( box, _bounding_boxes( i ) );
     }
 
     KOKKOS_INLINE_FUNCTION
     void join( volatile Box &dst, volatile Box const &src ) const
     {
-        for ( int d = 0; d < 3; ++d )
-        {
-            dst.minCorner()[d] =
-                KokkosHelpers::min( src.minCorner()[d], dst.minCorner()[d] );
-            dst.maxCorner()[d] =
-                KokkosHelpers::max( src.maxCorner()[d], dst.maxCorner()[d] );
-        }
+        expand( dst, src );
     }
 
   private:
--- a/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
+++ b/packages/Search/src/details/DTK_DetailsAlgorithms.hpp
@@ -114,9 +114,9 @@ void expand( Box &box, Sphere const &sphere )
     }
 }
 
-// check if two axis-aligned bounding boxes overlap
+// check if two axis-aligned bounding boxes intersect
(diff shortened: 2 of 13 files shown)


packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
```diff
@@ -144,8 +144,7 @@ KOKKOS_FUNCTION int nearestQuery( BVH<DeviceType> const bvh,
     {
         Node const *leaf = TreeTraversal<DeviceType>::getRoot( bvh );
         int const leaf_index = TreeTraversal<DeviceType>::getIndex( bvh, leaf );
-        double const leaf_distance =
-            distance( query_point, leaf->bounding_box );
+        double const leaf_distance = distance( leaf );
```
Rombur: I prefer the old notation. It is strange to compute the distance with only one input. If it wasn't for the diff, I would think that you are computing the distance between the leaf and the origin.

packages/Search/src/details/DTK_DetailsPredicate.hpp
```diff
@@ -24,115 +24,58 @@ struct SpatialPredicateTag
...
-    double _radius;
+    Geometry _geometry;
 };
 
-class Overlap
-{
-  public:
-    using Tag = SpatialPredicateTag;
+using Within = Intersects<Sphere>;
+using Overlap = Intersects<Box>;
```
Rombur: This two names are pretty bad. We should get ride of them. Within as nothing to do with being inside a volume and Overlap actually applies to both. The template names are a lot more meaningful.

packages/Search/src/details/DTK_DetailsAlgorithms.hpp
```diff
@@ -80,21 +95,36 @@ void expand( Box &box, Point const &point )
 }
 
 // expand an axis-aligned bounding box to include another box
+template <typename BOX,
+          typename = std::enable_if<std::is_same<
+              typename std::remove_volatile<BOX>::type, Box>::value>>
```
Rombur: Add a comment why you need to remove volatile

packages/Search/src/details/DTK_DetailsPredicate.hpp
dalg24: I agree with you.  I am trying to come up with better names.
In the meantime these aliases will do.

packages/Search/src/details/DTK_DetailsAlgorithms.hpp
```diff
@@ -73,6 +73,14 @@ double distance( Point const &point, Box const &box )
     return distance( point, projected_point );
 }
 
+// distance point-sphere
+KOKKOS_INLINE_FUNCTION
+double distance( Point const &point, Sphere const &sphere )
+{
+    return KokkosHelpers::max(
+        distance( point, sphere.centroid() ) - sphere.radius(), 0. );
```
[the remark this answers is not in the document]
dalg24: I am not sure how to call it.  This is consistent with the current implementation of `distance(point, box)`.

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
dalg24: `distance` here is a functor that measures the distance to nodes in the tree.\

Rombur: I know. It's just less easy to read than before.  
```C++
distance(query_point, leaf->bounding_box);
```
is clear that what you are computing while
```C++
distance(leaf);
```
is not

packages/Search/src/details/DTK_DetailsPredicate.hpp
sslattery: What is the Boost convention here?

sslattery: Actually do we even want a type alias here? Whats wrong with keeping them as `Intersects<T>`?

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
sslattery: So is the geometry for which we are querying distance now embedded in the distance functor?

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
```diff
@@ -205,15 +203,21 @@ KOKKOS_INLINE_FUNCTION int
...
 template <typename DeviceType, typename Predicate, typename Insert>
 KOKKOS_INLINE_FUNCTION int
 queryDispatch( BVH<DeviceType> const bvh, Predicate const &pred,
                Insert const &insert, NearestPredicateTag )
 {
-    return nearestQuery( bvh, pred._query_point, pred._k, insert );
+    auto const geometry = pred._geometry;
+    auto const k = pred._k;
+    return nearestQuery( bvh,
+                         [geometry]( Node const *node ) {
```
sslattery: Does this lambda have to be a `KOKKOS_LAMBDA`?

packages/Search/test/tstDetailsAlgorithms.cpp
```diff
@@ -31,58 +31,81 @@ TEUCHOS_UNIT_TEST( DetailsAlgorithms, distance )
...
+        dtk::intersects( box, {{{1.0, 0.0, 0.0}}, {{2.0, 1.0, 1.0}}} ) );
+    TEST_ASSERT(
+        dtk::intersects( box, {{{-0.5, -0.5, -0.5}}, {{0.5, 0.0, 0.5}}} ) );
+}
+
+TEUCHOS_UNIT_TEST( DetailsAlgorithms, intersects )
+{
+    DataTransferKit::Sphere sphere = {{{0., 0., 0.}}, 1.};
+    TEST_ASSERT( dtk::intersects( sphere, {{{0., 0., 0.}}, {{1., 1., 1.}}} ) );
+    TEST_ASSERT( !dtk::intersects( sphere, {{{1., 2., 3.}}, {{4., 5., 6.}}} ) );
```
sslattery: Do we have a sphere-sphere intersection or only sphere-box because we are checking intersection with tree nodes?

packages/Search/test/tstLinearBVH.cpp
```diff
@@ -636,16 +636,17 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( LinearBVH, rtree, DeviceType )
...
-    Kokkos::parallel_for( "register_nearest_queries",
-                          Kokkos::RangePolicy<ExecutionSpace>( 0, n_points ),
-                          KOKKOS_LAMBDA( int i ) {
-                              nearest_queries( i ) = details::nearest(
-                                  {{point_coords( i, 0 ), point_coords( i, 1 ),
-                                    point_coords( i, 2 )}},
-                                  k( i ) );
-                          } );
+    Kokkos::View<details::Nearest<DataTransferKit::Point> *, DeviceType>
+        nearest_queries( "neatest_queries", n_points );
```
sslattery: typo: `neatest_queries` -> `nearest_queries`

packages/Search/test/tstDistributedSearchTree.cpp
```diff
@@ -77,12 +77,14 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DistributedSearchTree, hello_world,
...
+    Kokkos::View<DataTransferKit::Details::Nearest<DataTransferKit::Point> *,
+                 DeviceType>
         nearest_queries( "nearest_queries", 1 );
     auto nearest_queries_host = Kokkos::create_mirror_view( nearest_queries );
-    nearest_queries_host( 0 ) = DataTransferKit::Details::nearest(
-        {{0.0 + comm_size - 1 - comm_rank, 0., 0.}},
-        comm_rank < comm_size - 1 ? 3 : 2 );
+    nearest_queries_host( 0 ) =
+        DataTransferKit::Details::nearest<DataTransferKit::Point>(
+            {{0.0 + comm_size - 1 - comm_rank, 0., 0.}},
```
sslattery: why add 0.0? Is this a way to cheat so it is for sure a floating point number?

packages/Search/src/details/DTK_DetailsAlgorithms.hpp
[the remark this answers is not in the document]
dalg24: No, this computes the distance to the sphere four points lying outside of it, and returns 0 for points inside.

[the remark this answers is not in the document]
Rombur: @sslattery Yes, which is not the distance to a sphere but to ball. The distance to a sphere is
```
|distance(point, centroid) - radius|
```
not
```
max(distance(point, centroid) - radius, 0)
```

packages/Search/src/details/DTK_DetailsPredicate.hpp
dalg24: Boost convention is to say the predicate type is "unspecified" and they use the convenience templated `intersects(geometry)` to generate the query (see [here])(http://www.boost.org/doc/libs/1_65_1/libs/geometry/doc/html/geometry/reference/spatial_indexes/group__predicates/intersects_geometry_const___.html#group__predicates_1ga78cb2ef221b951867c591ffb51b7d8c5).
We need to spell out the actual type of the queries because we make a Kokkos view of them.

packages/Search/test/tstDetailsAlgorithms.cpp
dalg24: This is sphrere-box intersection.  This is the only one needed to determine whether a sphere intersects with the nodes of the tree.  I did not bother writing sphere-sphere since it is not needed at this point.

packages/Search/test/tstDistributedSearchTree.cpp
dalg24: Yes.

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
dalg24: @sslattery Exactly.

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
dalg24: I don't think so since `queryDispatch()` is a `__host__ __device__` function.

packages/Search/src/details/DTK_DetailsAlgorithms.hpp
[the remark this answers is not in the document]
sslattery: ok that makes sense - so "sphere" is hollow and "ball" is solid

packages/Search/src/details/DTK_DetailsTreeTraversal.hpp
sslattery: so perhaps this can be resolved with a different naming convention like `distanceToGeometry` to infer that the geometry is in the functor

sslattery:
Anything else to do here before merging?

packages/Discretization/test/tstMeshGenerator.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "MeshGenerator.hpp"
 #include <Teuchos_UnitTestHarness.hpp>
+#include <fstream>
```
dalg24: @Rombur Got a compile error because that header was missing.  Piggybacking on that PR.

PR #340, opened 2017-12-12
Remove Shards dependency

Rombur:


packages/Discretization/src/DTK_PointInCell_def.hpp
```diff
@@ -31,140 +56,76 @@ void PointInCell<DeviceType>::search(
...
-
     // Perform the point in cell search. We hide the template parameters used by
     // Intrepid2, using the CellType template.
     // Note that if the Newton solver does not converge, Intrepid2 will just
     // return the last results and there is no way to know that the coordinates
     // in the reference frames where not found.
-    unsigned int const cell_topo_key = cell_topo.getKey();
-    if ( cell_topo_key ==
-         shards::getCellTopologyData<shards::Hexahedron<8>>()->key )
+    if ( cell_topo == DTK_HEX_8 )
```
dalg24: With `cell_topo` having an enumeration type, this actually becomes a great place for a `switch` statement.
You don't have to change though.

packages/Discretization/src/DTK_Topology.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include <Intrepid2_HGRAD_TRI_C1_FEM.hpp>
+#include <Intrepid2_HGRAD_TRI_C2_FEM.hpp>
+#include <Intrepid2_HGRAD_WEDGE_C1_FEM.hpp>
+#include <Intrepid2_HGRAD_WEDGE_C2_FEM.hpp>
+
+namespace DataTransferKit
+{
+struct Topology
+{
+  public:
```
dalg24: Unnecessary if you declare `Topology` to be a `struct`.

packages/Discretization/src/DTK_Topology.hpp
```diff
@@ -0,0 +1,153 @@
...
+        _topologies[DTK_WEDGE_6] = Topology( 3, 6, DTK_WEDGE_6 );
+        _topologies[DTK_WEDGE_15] = Topology( 3, 15, DTK_WEDGE_15 );
+        _topologies[DTK_WEDGE_18] = Topology( 3, 18, DTK_WEDGE_18 );
+    }
+
+    Topology &operator[]( int const i ) { return _topologies[i]; }
+    Topology const &operator[]( int const i ) const { return _topologies[i]; }
+
+  private:
+    std::array<Topology, DTK_N_TOPO> _topologies;
```
dalg24: Any reason why you chose an array over a map here?  (Just wondering)

packages/Discretization/src/DTK_PointInCell_def.hpp
```diff
@@ -12,16 +12,41 @@
 
 #include <DTK_DBC.hpp>
 #include <DTK_PointInCellFunctor.hpp>
+#include <DTK_Topology.hpp>
 #include <Kokkos_Core.hpp>
 
 namespace DataTransferKit
 {
+// Because search is static, we cannot use a private function so put the
+// function in its own namespace.
```
dalg24: I am not following you here.  Would you explain some more?

Rombur: The function `search` is static so it cannot call a `private` function because you would need to create the object.

packages/Discretization/src/DTK_Topology.hpp
Rombur: A map would work exactly the same way as the array but would be slower. The only reason it would be better is that  the key wouldn't have to be an `int` but I need it to be for a future PR.

PR #342, opened 2017-12-12
Fix unused parameter warnings

Rombur:


packages/Search/test/tstLinearBVH.cpp
```diff
@@ -515,7 +515,7 @@ std::vector<std::array<double, 3>> make_random_cloud( double Lx, double Ly,
 {
     std::vector<std::array<double, 3>> cloud( n );
     std::default_random_engine generator;
-    std::uniform_real_distribution<double> distribution_x( 0.0, Lz );
+    std::uniform_real_distribution<double> distribution_x( 0.0, Lx );
```
dalg24: I thought you got "all" of these...

packages/Utils/test/tstMiscellaneous.cpp
```diff
@@ -16,10 +16,10 @@
 TEUCHOS_UNIT_TEST( DataTransferKitRuntimeAPI, return_version )
 {
     auto const dtk_version = DataTransferKit::version();
-    TEUCHOS_ASSERT( !dtk_version.empty() );
+    TEST_ASSERT( !dtk_version.empty() );
```
dalg24: I had these changes sitting on some branch :)
Good catch.  Did you actually grep for `TEUCHOS_`?

Rombur:
Don't merge this yet. I will change the `configure` script so Jenkins can catch these warnings next time

packages/Search/test/tstLinearBVH.cpp
Rombur: me too :(

packages/Utils/test/tstMiscellaneous.cpp
Rombur: No, I got a warning that `out` and `success` were not used.

dalg24: Please do so and change if there is any other instance.

Rombur:
The problem is not the compiler but the flags that we are passing to gcc. These are not unused variables but unused parameters. We need `-Wextra` to get these warnings.

packages/Utils/test/tstMiscellaneous.cpp
Rombur: I just grepped and there shouldn't be any other `TEUCHOS_ASSERT`

Rombur:
I have fixed more unused parameter, enable shadow warnings, and added `Wextra` but only for clang so that we are not swamped with warnings when using the docker container.

Rombur:
Should be good to merge. We still have about 200 warnings split equally between Kokkos, Tpetra, and Teuchos.

PR #351, opened 2018-01-03
Improve Details::DistributedSearchTreeImpl::sortResults()

dalg24:
The goal is to enable reuse in interpolation operator.

### DIFF
--- a/packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
+++ b/packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
@@ -80,11 +80,9 @@ struct DistributedSearchTreeImpl
                                Kokkos::View<int *, DeviceType> &indices,
                                Kokkos::View<int *, DeviceType> &offset,
                                Kokkos::View<int *, DeviceType> &ranks );
-    static void
-    sortResults( Kokkos::View<int *, DeviceType> query_ids,
-                 Kokkos::View<int *, DeviceType> results,
-                 Kokkos::View<int *, DeviceType> ranks,
-                 Kokkos::View<double *, DeviceType> *distances_ptr = nullptr );
+
+    template <typename View, typename... OtherViews>
+    static void sortResults( View keys, OtherViews... other_views );
 
     static void countResults( int n_queries,
                               Kokkos::View<int *, DeviceType> query_ids,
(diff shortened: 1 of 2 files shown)


packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -258,38 +256,46 @@ void DistributedSearchTreeImpl<DeviceType>::queryDispatch(
     ////////////////////////////////////////////////////////////////////////////
 }
 
+// NOTE: for some reason Kokkos::BinSort::sort() is not const...
```
dalg24: I added the `const` specifier in `kokkos/algorithms/src/Kokkos_Sort.hpp` and it seemed to compile just fine.
It might be worth opening an issue on the Kokkos side.

dalg24: c.f. kokkos/kokkos#1308

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -258,38 +256,46 @@ void DistributedSearchTreeImpl<DeviceType>::queryDispatch(
...
     if ( result.min_val == result.max_val )
         return;
-    Kokkos::BinSort<Kokkos::View<int *, DeviceType>, CompType> bin_sort(
-        query_ids, CompType( n / 2, result.min_val, result.max_val ), true );
+    Kokkos::BinSort<View, Comp> bin_sort(
+        keys, Comp( n / 2, result.min_val, result.max_val ), true );
     bin_sort.create_permute_vector();
-    bin_sort.sort( results );
-    bin_sort.sort( ranks );
-    if ( distances_ptr )
```
Rombur: Why don't you need to check this anymore?

dalg24: Because now I pass the view rather than a pointer to it.  The old signature with a default value `nullptr` for the pointer was a hack to not implement two version of `sortResults()`.

dalg24:
@Rombur Kokkos won't throw if `BinSort::sort()` is called on a view that does not have the same extent as the keys.  I added a precondition check in DTK.

Rombur:
ok thanks

Rombur:
Yeah squash and merge yourself

PR #355, opened 2018-01-12
Communication helpers

dalg24:
* Updated `DistributedSearchTreeImpl::sendAcrossNetwork()` to handle view with rank > 1
* Added `Details::NearestNeighborOperatorImpl::fetch()` that will also be used in the MLS operator

### DIFF
--- a/packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
+++ b/packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
@@ -28,7 +28,7 @@ struct NearestNeighborOperatorImpl
     // boxes in their constructors.
     static DistributedSearchTree<DeviceType> makeDistributedSearchTree(
         Teuchos::RCP<const Teuchos::Comm<int>> const &comm,
-        Kokkos::View<Coordinate **, DeviceType> const &source_points )
+        Kokkos::View<Coordinate const **, DeviceType> source_points )
     {
         int const n_source_points = source_points.extent( 0 );
         Kokkos::View<Box *, DeviceType> boxes( "boxes", n_source_points );
@@ -46,7 +46,7 @@ struct NearestNeighborOperatorImpl
 
     static Kokkos::View<Details::Nearest<DataTransferKit::Point> *, DeviceType>
     makeNearestNeighborQueries(
-        Kokkos::View<Coordinate **, DeviceType> const &target_points )
+        Kokkos::View<Coordinate const **, DeviceType> target_points )
     {
         int const n_target_points = target_points.extent( 0 );
         Kokkos::View<Details::Nearest<DataTransferKit::Point> *, DeviceType>
--- a/packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
+++ b/packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
@@ -63,13 +63,17 @@ struct NearestNeighborOperatorImpl
         return nearest_queries;
     }
 
+    template <typename View>
     static void
     pullSourceValues( Teuchos::RCP<const Teuchos::Comm<int>> const &comm,
-                      Kokkos::View<double *, DeviceType> source_values,
+                      View source_values,
                       Kokkos::View<int *, DeviceType> &buffer_indices,
                       Kokkos::View<int *, DeviceType> &buffer_ranks,
-                      Kokkos::View<double *, DeviceType> &buffer_values )
+                      typename View::non_const_type &buffer_values )
     {
+        static_assert(
+            View::rank <= 2,
+            "pullSourceValues() requires rank-1 or rank-2 view arguments" );
         int const n_exports = buffer_indices.extent( 0 );
         Tpetra::Distributor distributor( comm );
         int const n_imports =
@@ -99,32 +103,37 @@ struct NearestNeighborOperatorImpl
 
         buffer_indices = import_target_indices;
         buffer_ranks = import_ranks;
-        Kokkos::realloc( buffer_values, n_imports );
+        Kokkos::realloc( buffer_values, n_imports,
+                         source_values.dimension_1() );
         Kokkos::parallel_for(
             DTK_MARK_REGION( "get_source_values" ),
             Kokkos::RangePolicy<ExecutionSpace>( 0, n_imports ),
             KOKKOS_LAMBDA( int i ) {
-                buffer_values( i ) =
-                    source_values( import_source_indices( i ) );
+                for ( int j = 0; j < (int)source_values.dimension_1(); ++j )
+                    buffer_values( i, j ) =
+                        source_values( import_source_indices( i ), j );
             } );
         Kokkos::fence();
     }
 
+    template <typename View>
     static void
     pushTargetValues( Teuchos::RCP<const Teuchos::Comm<int>> const &comm,
                       Kokkos::View<int *, DeviceType> const &buffer_indices,
                       Kokkos::View<int *, DeviceType> const &buffer_ranks,
-                      Kokkos::View<double *, DeviceType> const &buffer_values,
-                      Kokkos::View<double *, DeviceType> target_values )
+                      View const &buffer_values, View target_values )
     {
+        static_assert(
+            View::rank <= 2,
+            "pushTargetValues() requires rank-1 or rank-2 view arguments" );
         Tpetra::Distributor distributor( comm );
         int const n_imports =
             distributor.createFromSends( Teuchos::ArrayView<int>(
                 buffer_ranks.data(), buffer_ranks.extent( 0 ) ) );
 
-        Kokkos::View<double *, DeviceType> export_source_values = buffer_values;
-        Kokkos::View<double *, DeviceType> import_source_values(
-            "source_values", n_imports );
+        View export_source_values = buffer_values;
+        View import_source_values( "source_values", n_imports,
+                                   target_values.dimension_1() );
         DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
             distributor, export_source_values, import_source_values );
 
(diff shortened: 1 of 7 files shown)


packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
```diff
@@ -138,11 +147,39 @@ struct NearestNeighborOperatorImpl
...
             } );
         Kokkos::fence();
     }
+
+    template <typename View>
+    static typename View::non_const_type
+    fetch( Teuchos::RCP<Teuchos::Comm<int> const> const &comm,
+           Kokkos::View<int const *, DeviceType> ranks,
+           Kokkos::View<int const *, DeviceType> indices, View values )
+    {
```
dalg24: Preconditions?

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -106,26 +106,79 @@ struct DistributedSearchTreeImpl
...
-    // should be able to directly call doPostsAndWaits() on the views passed as
-    // argument.  See https://github.com/trilinos/Trilinos/issues/1454
-    auto exports_host = Kokkos::create_mirror_view( exports );
+    DTK_REQUIRE( ( exports.dimension_1() == imports.dimension_1() ) &&
+                 ( exports.dimension_2() == imports.dimension_2() ) &&
+                 ( exports.dimension_3() == imports.dimension_3() ) &&
+                 ( exports.dimension_4() == imports.dimension_4() ) &&
+                 ( exports.dimension_5() == imports.dimension_5() ) &&
+                 ( exports.dimension_6() == imports.dimension_6() ) &&
+                 ( exports.dimension_7() == imports.dimension_7() ) );
```
dalg24: Make sure to trigger exception in test.  Also might want to check that distributor and view sizes are compatible.

packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
dalg24: OK.  Added pre and post conditions check.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
dalg24: Added check for compatible distributor and exports/imports views.

dalg24: Didn't bother adding a test to trigger though.

dalg24:
Why d411a9e passed and 6ec8887 failed is a mystery to me.  Any idea?

Rombur:
The failing test is running for 25 minutes, so it looks like a dead lock. Is is possible that the condition is not verified on one processor, that this proc throws, and then, you deadlock when you need to communicate?

dalg24:
3a381b3 does pass the tests.

dalg24:
Checks in `fetch()` were flat out wrong...  Fixed now.

packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
```diff
@@ -138,11 +147,39 @@ struct NearestNeighborOperatorImpl
...
+                for ( int j = 0; j < (int)target_values.dimension_1(); ++j )
+                    target_values( import_target_indices( i ), j ) =
+                        import_source_values( i, j );
             } );
         Kokkos::fence();
     }
+
+    template <typename View>
+    static typename View::non_const_type
+    fetch( Teuchos::RCP<Teuchos::Comm<int> const> const &comm,
```
Rombur: I am confused on what this function does. Why does fetch does a `pull` and a `push`? `fetch` for me means that it only does a `pull`

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <Tpetra_Distributor.hpp>
+
+template <
+    typename View,
+    typename std::enable_if<
+        std::is_same<
+            typename View::traits::memory_space,
+            typename View::traits::host_mirror_space::memory_space>::value,
+        int>::type = 0>
+Teuchos::ArrayView<typename View::const_value_type> toArray( View const &v )
```
Rombur: Why is this function `const_value_type` and the other `non_const`? Shouldn't keep the constness (const View returns const array and non const View returns non const array)?

Rombur: Nevermind I didn't see that this only a helper for the test.

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
```diff
@@ -0,0 +1,172 @@
...
+        Teuchos::DefaultComm<int>::getComm();
+    int const comm_rank = comm->getRank();
+    int const comm_size = comm->getSize();
+
+    int const DIM = 3;
+
+    // send 1 packet to rank k
+    // receive comm_size packets
+    Kokkos::View<int **, DeviceType> u_exp( "u_exp", comm_size, DIM );
+    Kokkos::parallel_for( "",
```
Rombur: I don't think that you have to give a name if you don't want to. BTW I am fine with not giving any name in test.

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
```diff
@@ -155,6 +168,63 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DetailsDistributedSearchTreeImpl,
...
+
+    Teuchos::RCP<const Teuchos::Comm<int>> comm =
+        Teuchos::DefaultComm<int>::getComm();
+    int const comm_rank = comm->getRank();
+    int const comm_size = comm->getSize();
+
+    // make communicaton plan
+    Kokkos::View<int *, DeviceType> indices( "indices", comm_size );
+    Kokkos::View<int *, DeviceType> ranks( "ranks", comm_size );
+    Kokkos::parallel_for( "fill_indices_and_ranks",
```
Rombur: Maybe use `DTK_MARK_REGION` you didn't do in the test. It's up to you, I don't think it's important in the test.

packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
dalg24: `pullSourceValues()` and `pushTargetValues()` are implementation details. `fetch()` allow you to retrieve values on a distributed array.

packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
```diff
@@ -138,11 +147,39 @@ struct NearestNeighborOperatorImpl
...
+            Kokkos::create_mirror( DeviceType(), indices );
+        Kokkos::deep_copy( buffer_indices, indices );
+
+        typename View::non_const_type buffer_values( values.label() );
+
+        pullSourceValues( comm, values, buffer_indices, buffer_ranks,
+                          buffer_values );
+
+        typename View::non_const_type values_out(
+            values.label(), ranks.extent( 0 ), values.dimension_1() );
```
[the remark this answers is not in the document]
dalg24: No reason, I will change it to `values.extent( 1 )`.

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
dalg24: I forgot about the version of `parallel_for()` that does not take a label.  I will change it.

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
dalg24: I was only providing a label because I thought I had to.  I will either remove or leave as is.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -106,6 +106,48 @@ struct DistributedSearchTreeImpl
 template <typename DeviceType>
 double DistributedSearchTreeImpl<DeviceType>::epsilon = 1.0e-6;
 
+template <typename View>
+inline Kokkos::View<typename View::traits::non_const_data_type,
```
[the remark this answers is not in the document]
dalg24: No good reason for casting away the `const`.  I will change.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -106,6 +106,48 @@ struct DistributedSearchTreeImpl
...
+create_layout_right_mirror_view(
+    View const &src,
+    typename std::enable_if<(
+        ( std::is_same<typename View::traits::array_layout,
+                       Kokkos::LayoutRight>::value ||
+          View::rank == 1 && !std::is_same<typename View::traits::array_layout,
+                                           Kokkos::LayoutStride>::value ) &&
+        std::is_same<typename View::traits::memory_space,
+                     typename View::traits::host_mirror_space::memory_space>::
+            value )>::type * = 0 )
```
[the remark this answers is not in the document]
dalg24: It is correct, I meant to use the precedence of the AND on the OR.

if ( ( layout is right OR ( rank is one AND layout is not strided ) ) AND ... )

dalg24:
@Rombur I pushed changes to address your comments.  I will squash them before the merge.  Let me know if you are happy with the current version.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -106,26 +108,86 @@ struct DistributedSearchTreeImpl
...
+    typename std::enable_if<!(
+        ( std::is_same<typename View::traits::array_layout,
+                       Kokkos::LayoutRight>::value ||
+          View::rank == 1 && !std::is_same<typename View::traits::array_layout,
+                                           Kokkos::LayoutStride>::value ) &&
+        std::is_same<typename View::traits::memory_space,
+                     typename View::traits::host_mirror_space::memory_space>::
+            value )>::type * = 0 )
+{
+    return Kokkos::View<typename View::traits::non_const_data_type,
```
Rombur: I think that you forgot to change this line. You cast away the const but it doesn't match the return type of the function

dalg24: Good catch.  Fixed.

PR #356, opened 2018-01-18
Add constant polynomial basis

dalg24:


packages/Meshfree/src/DTK_MultivariatePolynomialBasis.hpp
```diff
@@ -95,6 +114,15 @@ KOKKOS_INLINE_FUNCTION
              p[1] * p[1], p[1] * p[2], p[2] * p[2]}};
 }
 
+template <>
+template <typename Point>
+KOKKOS_INLINE_FUNCTION
+    Kokkos::Array<double, MultivariatePolynomialBasis<Constant, 2>::size()>
+    MultivariatePolynomialBasis<Constant, 2>::operator()( Point const &p ) const
+{
+    return {{1.}};
```
Rombur: Can't you do partial specialization do avoid the the copy/paste?

dalg24: I failed miserably when I tried.  Can't do partial specialization on member function.

PR #358, opened 2018-01-22
Add CUDA 9.1 pkg version to setup script for nvidia-docker

dalg24:
@sslattery please see if this works on your system


Rombur:
Don't forget that you won't be able to compile Trilinos with CUDA 9.1. You need the development version of Kokkos to be able to compile Intrepid2 (and the latest version of develop broke something in Trilinos so you need to wait for the fix to be merged in Trilinos).

dalg24:
@Rombur It has nothing to do with the CUDA version you use when you configure DTK/Trilinos.  This has only to do with making development/testing containers CUDA-aware.

Rombur:
I know

sslattery:
Tested and confirmed I was able to extend and build the image with the 9.1 drivers on my system. Explicitly setting `CUDA_VERSION=8.0` also gave a successful build.

PR #361, opened 2018-01-24
Implement "exact" nearest neighbors search on distributed tree

dalg24:
Previous implementation would only yield an "approximate" nearest neighbors.  It was leveraging `Details::DistributedSearchTreeImpl::epsilon` to produce an algorithm equivalent to what was in DTK-2.0.

Note that `DistributedSearchTreeImpl` was moved into `Details::` namespace which will affect you if you use `sendAcrossNetwork()`.  I plan on moving this functionality, along with `fetch()`, into some communication helpers `struct` soon.

### DIFF
--- a/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
+++ b/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
@@ -18,8 +18,8 @@
 #include <Teuchos_RCP.hpp>
 
 #include <DTK_DBC.hpp>
+#include <DTK_DetailsDistributedSearchTreeImpl.hpp>
 #include <DTK_LinearBVH.hpp>
-#include <details/DTK_DetailsDistributedSearchTreeImpl.hpp>
 
 #include "DTK_ConfigDefs.hpp"
 
@@ -47,7 +47,7 @@ class DistributedSearchTree
     using SizeType = typename BVH<DeviceType>::SizeType;
     /** Returns the global number of objects stored in the tree.
      */
-    inline SizeType size() const { return _size; }
+    inline SizeType size() const { return _top_tree_size; }
 
     /** Indicates whether the tree is empty on all processes.
      */
@@ -99,7 +99,8 @@ class DistributedSearchTree
     Teuchos::RCP<Teuchos::Comm<int> const> _comm;
     BVH<DeviceType> _top_tree;    // replicated
     BVH<DeviceType> _bottom_tree; // local
-    SizeType _size;
+    SizeType _top_tree_size;
+    Kokkos::View<SizeType *, DeviceType> _bottom_tree_sizes;
 };
 
 template <typename DeviceType>
--- a/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
+++ b/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
@@ -42,7 +42,7 @@ class DistributedSearchTree
     /** Returns the smallest axis-aligned box able to contain all the objects
      *  stored in the tree or an invalid box if the tree is empty.
      */
-    inline Box bounds() const { return _distributed_tree.bounds(); }
+    inline Box bounds() const { return _top_tree.bounds(); }
 
     using SizeType = typename BVH<DeviceType>::SizeType;
     /** Returns the global number of objects stored in the tree.
@@ -97,8 +97,8 @@ class DistributedSearchTree
 
   private:
     Teuchos::RCP<Teuchos::Comm<int> const> _comm;
-    BVH<DeviceType> _local_tree;
-    BVH<DeviceType> _distributed_tree;
+    BVH<DeviceType> _top_tree;    // replicated
+    BVH<DeviceType> _bottom_tree; // local
     SizeType _size;
 };
 
@@ -112,7 +112,7 @@ void DistributedSearchTree<DeviceType>::query(
 {
     using Tag = typename Query::Tag;
     DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _distributed_tree, _local_tree, queries, indices, offset, ranks,
+        _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks,
         Tag{} );
 }
 
@@ -130,8 +130,8 @@ DistributedSearchTree<DeviceType>::query(
 {
     using Tag = typename Query::Tag;
     DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _distributed_tree, _local_tree, queries, indices, offset, ranks,
-        Tag{}, &distances );
+        _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks, Tag{},
+        &distances );
 }
 
 } // end namespace DataTransferKit
--- a/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
+++ b/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
@@ -112,7 +112,7 @@ void DistributedSearchTree<DeviceType>::query(
     Kokkos::View<int *, DeviceType> &ranks ) const
 {
     using Tag = typename Query::Tag;
-    DistributedSearchTreeImpl<DeviceType>::queryDispatch(
+    Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
         _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks,
         Tag{} );
 }
@@ -130,7 +130,7 @@ DistributedSearchTree<DeviceType>::query(
     Kokkos::View<double *, DeviceType> &distances ) const
 {
     using Tag = typename Query::Tag;
-    DistributedSearchTreeImpl<DeviceType>::queryDispatch(
+    Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
         _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks, Tag{},
         &distances );
 }
--- a/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
+++ b/packages/Search/src/DTK_DistributedSearchTree_decl.hpp
@@ -96,6 +96,7 @@ class DistributedSearchTree
            Kokkos::View<double *, DeviceType> &distances ) const;
 
   private:
+    friend struct Details::DistributedSearchTreeImpl<DeviceType>;
     Teuchos::RCP<Teuchos::Comm<int> const> _comm;
     BVH<DeviceType> _top_tree;    // replicated
     BVH<DeviceType> _bottom_tree; // local
@@ -113,8 +114,7 @@ void DistributedSearchTree<DeviceType>::query(
 {
     using Tag = typename Query::Tag;
     Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks,
-        Tag{} );
+        *this, queries, indices, offset, ranks, Tag{} );
 }
 
 template <typename DeviceType>
@@ -131,8 +131,7 @@ DistributedSearchTree<DeviceType>::query(
 {
     using Tag = typename Query::Tag;
     Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks, Tag{},
-        &distances );
+        *this, queries, indices, offset, ranks, Tag{}, &distances );
 }
 
 } // end namespace DataTransferKit
--- a/packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
+++ b/packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
@@ -34,8 +34,8 @@ struct DistributedSearchTreeImpl
     // spatial queries
     template <typename Query>
     static void queryDispatch( Teuchos::RCP<Teuchos::Comm<int> const> comm,
-                               BVH<DeviceType> const &distributed_tree,
-                               BVH<DeviceType> const &local_tree,
+                               BVH<DeviceType> const &top_tree,
+                               BVH<DeviceType> const &bottom_tree,
                                Kokkos::View<Query *, DeviceType> queries,
                                Kokkos::View<int *, DeviceType> &indices,
                                Kokkos::View<int *, DeviceType> &offset,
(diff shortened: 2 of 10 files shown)


Rombur:
From Travis ```Searching for do/scratch/packages/Search/src/DTK_DistributedSearchTree_decl.hpp:101: warning: documented symbol `DistributedSearchTreeImpl< DeviceType >' was not declared or defined.```

dalg24:
Yeah, I saw.  I will try to add a forward declaration to make Doxygen happy.

dalg24:
I am really not sure why Doxygen says the symbol has not been declared nor defined.  I checked and the file `DTK_DetailsDistributedSearchTreeImpl.hpp` does get parsed.  I am tempted to use `#ifndef DOXYGEN_SHOULD_SKIP_THIS` since it is implementation details anyway.  What do you think Bruno?

dalg24:
Nevermind, it seems that adding `struct` in the friend declaration do the trick.

Rombur:
great

packages/Search/test/tstDistributedSearchTree.cpp
```diff
@@ -297,6 +297,62 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DistributedSearchTree, one_leaf_per_rank,
...
+                {{{(double)( comm_size - 1 - comm_rank ) + .75, 0., 0.}}, 1},
+            } ),
+            {0}, {0, 1}, {comm_size - comm_rank}, success, out );
+    else
+        checkResults(
+            tree,
+            makeNearestQueries<DeviceType>( {
+                {{{(double)( comm_size - 1 - comm_rank ) + .75, 0., 0.}}, 1},
+            } ),
+            {0}, {0, 1}, {comm_size - 1}, success, out );
```
Rombur: Why are the queries on rank 0 and 1 have the same reference point?

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
```diff
@@ -97,8 +97,8 @@ class DistributedSearchTree
 
   private:
     Teuchos::RCP<Teuchos::Comm<int> const> _comm;
-    BVH<DeviceType> _local_tree;
-    BVH<DeviceType> _distributed_tree;
+    BVH<DeviceType> _top_tree;    // replicated
+    BVH<DeviceType> _bottom_tree; // local
```
Rombur: If I didn't know what `bottom_tree` and `top_tree` were, there is no way I would understand what they are.

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
```diff
@@ -115,8 +116,7 @@ void DistributedSearchTree<DeviceType>::query(
 {
     using Tag = typename Query::Tag;
     Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _top_tree, _bottom_tree, queries, indices, offset, ranks,
-        Tag{} );
+        *this, queries, indices, offset, ranks, Tag{} );
```
Rombur: If you need to make `DistributedSearchTreeImpl` a friend and pass `*this`, why not move `queryDispatch` to the class. Because this is basically what you are doing but in a complicated way.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -57,11 +57,19 @@ struct DistributedSearchTreeImpl
         Kokkos::View<double *, DeviceType> *distances_ptr = nullptr );
 
     template <typename Query>
-    static void deviseStrategy( Point epsilon,
-                                Kokkos::View<Query *, DeviceType> queries,
+    static void deviseStrategy( Kokkos::View<Query *, DeviceType> queries,
```
Rombur: Please add a comment on what `deviseStrategy` and `reassessStrategy` do. It hasn't been that long and yet I don't have any clue on what `deviseStrategy` is supposed to do.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -198,36 +206,96 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
 
 template <typename DeviceType>
 template <typename Query>
-void DistributedSearchTreeImpl<DeviceType>::deviseStrategy(
-    Point eps, Kokkos::View<Query *, DeviceType> queries,
+void DistributedSearchTreeImpl<DeviceType>::reassessStrategy(
```
Rombur: maybe implement the functions in the order you have declared them. This way it is easier to find them

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -198,36 +206,96 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
...
+    Kokkos::View<int *, DeviceType> &offset,
+    Kokkos::View<double *, DeviceType> & )
+{
+    auto const &top_tree = tree._top_tree;
+    auto const &bottom_tree_sizes = tree._bottom_tree_sizes;
+
+    top_tree.query( queries, indices, offset );
+
+    auto const n_queries = queries.extent( 0 );
+    Kokkos::View<int *, DeviceType> _offset( offset.label(), n_queries + 1 );
```
Rombur: Is this a `private` variable?

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -198,36 +206,96 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
...
+    Kokkos::parallel_for(
+        DTK_MARK_REGION( "bottom_trees_with_required_cumulated_leaves_count" ),
+        Kokkos::RangePolicy<ExecutionSpace>( 0, n_queries ),
+        KOKKOS_LAMBDA( int i ) {
+            int leaves_count = 0;
+            for ( int j = offset( i ); j < offset( i + 1 ); ++j )
+            {
+                leaves_count += bottom_tree_sizes[indices( j )];
+                ++_offset( i );
+                if ( leaves_count >= queries( i )._k )
```
Rombur: What does this condition mean?

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -198,36 +206,96 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
...
+                    break;
+            }
+        } );
+    Kokkos::fence();
+
+    exclusivePrefixSum( _offset );
+
+    Kokkos::View<int *, DeviceType> _indices( indices.label(),
```
Rombur: Is this a `private` variable?

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -240,48 +308,68 @@ void DistributedSearchTreeImpl<DeviceType>::queryDispatch(
...
-    // Merge results
-    ////////////////////////////////////////////////////////////////////////////
-    int const n_queries = queries.extent_int( 0 );
-    countResults( n_queries, ids, offset );
-    sortResults( ids, indices, ranks, distances );
-    filterResults( queries, distances, indices, offset, ranks );
-    ////////////////////////////////////////////////////////////////////////////
+    // "Strategy" is used to determine what ranks to forward queries to.  In
+    // the 1st pass, the queries are sent to as many ranks as necessary to
+    // guaratee that all k neighbors queried for are found.  In the 2nd pass,
```
Rombur: typo `guaratee`

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -641,7 +641,7 @@ void DistributedSearchTreeImpl<DeviceType>::filterResults(
 
     exclusivePrefixSum( _offset );
 
-    int const n_truncated_results = _offset( n_queries );
+    int const n_truncated_results = lastElement( _offset );
```
Rombur: good catch

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
```diff
@@ -111,9 +114,8 @@ void DistributedSearchTree<DeviceType>::query(
     Kokkos::View<int *, DeviceType> &ranks ) const
 {
     using Tag = typename Query::Tag;
-    DistributedSearchTreeImpl<DeviceType>::queryDispatch(
-        _comm, _distributed_tree, _local_tree, queries, indices, offset, ranks,
-        Tag{} );
+    Details::DistributedSearchTreeImpl<DeviceType>::queryDispatch(
+        *this, queries, indices, offset, ranks, Tag{} );
```
Rombur: If you need to make `DistributedSearchTreeImpl `a friend and pass `*this`, why not move `queryDispatch` to the class. Because this is basically what you are doing but in a complicated way.

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
```diff
@@ -96,10 +98,11 @@ class DistributedSearchTree
            Kokkos::View<double *, DeviceType> &distances ) const;
 
   private:
+    friend Details::DistributedSearchTreeImpl<DeviceType>;
     Teuchos::RCP<Teuchos::Comm<int> const> _comm;
-    BVH<DeviceType> _local_tree;
-    BVH<DeviceType> _distributed_tree;
-    SizeType _size;
+    BVH<DeviceType> _top_tree;    // replicated
+    BVH<DeviceType> _bottom_tree; // local
```
Rombur: If I didn't know what `bottom_tree` and `top_tree` were, there is no way I would understand what they are.

packages/Search/test/tstDistributedSearchTree.cpp
dalg24: You mean same nearest neighbor?  See the ascii drawing above.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
dalg24: Will swap order in the definitions.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
dalg24: That the ranks we will forward queries to have a cumulated total number of leaves big enough to find all k neighbors queried for.

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
dalg24: Another motivation for that was, if I make it a member of the class, I will not be able to mark it as private if I want to use Kokkos lambdas.

packages/Search/src/DTK_DistributedSearchTree_decl.hpp
Rombur: Did we put anywhere what the trees are? I am just worry that if someone is new on the project there is no way to know that these are

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -197,99 +199,192 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
...
+    Kokkos::parallel_for(
+        DTK_MARK_REGION( "bottom_trees_with_required_cumulated_leaves_count" ),
+        Kokkos::RangePolicy<ExecutionSpace>( 0, n_queries ),
+        KOKKOS_LAMBDA( int i ) {
+            int leaves_count = 0;
+            for ( int j = offset( i ); j < offset( i + 1 ); ++j )
+            {
+                leaves_count += bottom_tree_sizes( indices( j ) );
+                ++_offset( i );
+                if ( leaves_count >= queries( i )._k )
```
Rombur: Is there anything we could/should do if there is not enough leaves?

dalg24: Now that I think about if we could also break if `bottom_tree_sizes( indices( j ) == 0 `.  No need to forward the query to empty trees.

Besides that no, not much you can do.  `query()` guarantees that if you have neighbors and it will find them.  The results that are returned are all valid entries, in other words, if you query for 10 neighbors and they are only 2 leaves, it will only returned the two indices sorted from the nearest to the farthest.

Rombur: What I was thinking but it's not clear from what I wrote, is what happens if I asked for 10 neighbors but you have only 8 points total? I guess it shouldn't happen in a real case but if someone has a toy problem they could create such a case.

dalg24: We are supposed to handle such cases.  I just looked at `tstDistributedSearchTree.cpp` and it might be a good occasion to add some test cases.

In short, say you only have one leave on rank 0, and that all other ranks have empty local trees.
Then if you query for 2 nearest neighbors (from any rank), you should expect offset={0,1}, indices={0}, ranks={0}

Rombur: ok thanks

dalg24: I added a couple tests.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
dalg24: I renamed these.

packages/Search/src/details/DTK_DetailsDistributedSearchTreeImpl.hpp
```diff
@@ -197,99 +199,197 @@ DistributedSearchTreeImpl<DeviceType>::sendAcrossNetwork(
...
+    Kokkos::View<int *, DeviceType> &offset,
+    Kokkos::View<double *, DeviceType> & )
 {
-    int const n_queries = queries.extent( 0 );
-    Kokkos::View<Details::Overlap *, DeviceType> overlap_queries(
-        "overlap_queries", n_queries );
+    auto const &top_tree = tree._top_tree;
+    auto const &bottom_tree_sizes = tree._bottom_tree_sizes;
+
+    // Find the k nearest local trees.
```
Rombur: You should grep for `local` and `distributed` and replace them by `bottom` and `top`

dalg24: Comment is accurate.  I had already renamed all variables.

Rombur: I don't understand. The comments here and below are about local trees but you have renamed them bottom trees, e.g., `Stop if local trees get empty` but the code is `if ( bottom_tree_size == 0 ) `

PR #362, opened 2018-01-29
Add isValid(geometry)

dalg24:


packages/Search/src/details/DTK_DetailsAlgorithms.hpp
```diff
@@ -45,6 +45,30 @@ bool equals( Sphere const &l, Sphere const &r )
...
+    for ( int d = 0; d < 3; ++d )
+        if ( !isFinite( p[d] ) )
+            return false;
+    return true;
+}
+
+KOKKOS_INLINE_FUNCTION
+bool isValid( Box const &b )
+{
+    return isValid( b.minCorner() ) && isValid( b.maxCorner() );
```
Rombur: Shouldn't we also check that `minCorner` is less or equal than `maxCorner`?

packages/Search/test/tstDetailsAlgorithms.cpp
```diff
@@ -152,3 +152,29 @@ TEUCHOS_UNIT_TEST( DetailsAlgorithms, centroid )
...
+    TEST_ASSERT( !isValid( Point{{0., infty, 0.}} ) );
+
+    TEST_ASSERT( isValid( Box{{{1., 2., 3.}}, {{4., 5., 6.}}} ) );
+    TEST_ASSERT( isValid( Box{{{0., 0., 0.}}, {{0., 0., 0.}}} ) );
+    TEST_ASSERT( !isValid( Box{{{0., 0., -infty}}, {{0., 0., 0.}}} ) );
+    TEST_ASSERT( !isValid( Box{{{0., 0., 0.}}, {{+infty, 0., 0.}}} ) );
+    TEST_ASSERT( isValid( Box{} ) );
+
+    TEST_ASSERT( isValid( Sphere{{{1., 2., 3.}}, 4.} ) );
+    TEST_ASSERT( !isValid( Sphere{{{0., 0., 0.}}, 0.} ) );
```
Rombur: Why is this valid for a `Box` but not for a `Sphere`?

packages/Search/test/tstDetailsAlgorithms.cpp
```diff
@@ -152,3 +152,29 @@ TEUCHOS_UNIT_TEST( DetailsAlgorithms, centroid )
...
+    TEST_ASSERT( !isValid( Sphere{{{1., 2., 3.}}, -1.} ) );
+    TEST_ASSERT( !isValid( Sphere{{{0., 0., 0.}}, +infty} ) );
+    TEST_ASSERT( !isValid( Sphere{{{0., -infty, 0.}}, +1.} ) );
+    TEST_ASSERT( !isValid( Sphere{} ) );
```
Rombur: Why is this valid for a `Box` but not a `Sphere`?

packages/Search/test/tstDetailsAlgorithms.cpp
dalg24: radius has to greater than zero

packages/Search/test/tstDetailsAlgorithms.cpp
dalg24: default constructor initialize radius to zero

packages/Search/test/tstDetailsAlgorithms.cpp
Rombur: Why?

dalg24: We can change to greater than or to zero, that's fine with me

Rombur: I think it makes sense to have the same behavior for `Box` and `Sphere`. If I can build a `Box` of size zero, I should be able to build a `Sphere` of radius zero (unless we have a good reason).

packages/Search/src/details/DTK_DetailsAlgorithms.hpp
dalg24: That would invalidate "empty" box returned by default constructor.

packages/Search/test/tstDetailsAlgorithms.cpp
dalg24: Agreed.  Will change.

PR #364, opened 2018-01-31
Amend to #336

dalg24:
#361 introduced a couple changes that were not reflected in #336 when it got merged into master:
* `DistributedSearchTreeImpl::epsilon` was removed.  (The nearest neighbors search on the distributed tree does not require this "knob" any more to guarantee the exact result.)
* `DistributedSearchTreeImpl` was moved into `Details::` namespace

This could have been avoided, had the PR been rebased onto master.  Oh well...

### DIFF
--- a/packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
+++ b/packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
@@ -163,13 +163,6 @@ void testUniquelyOwnedProblem(
     using ExecutionSpace = typename Device::execution_space;
     using Scalar = double;
 
-    // Set an epsilon for the search. The meshes (see below) mesh a sphere of
-    // radius 10. An epsilon of 1.0 is needed to fully capture nearest
-    // neighbors in adjacent partitions.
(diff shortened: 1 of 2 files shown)


sslattery:
My bad...

It would be nice to get slack notifications or even emails when the master build fails. Is there an easy way to do that?

packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -163,13 +163,6 @@ void testUniquelyOwnedProblem(
     using ExecutionSpace = typename Device::execution_space;
     using Scalar = double;
 
-    // Set an epsilon for the search. The meshes (see below) mesh a sphere of
```
sslattery: So we no longer need an epsilon because we now have exact nearest neighbors - right?

dalg24: Correct.

dalg24:
> It would be nice to get slack notifications or even emails when the master build fails.

We can do both.  I am looking into that.

PR #365, opened 2018-01-31
Fix clang warnings from #336

dalg24:


packages/Meshfree/test/tstNearestNeighborExodusGenerator.cpp
```diff
@@ -156,12 +156,18 @@ void testUniquelyOwnedProblem(
     const Kokkos::View<double **, ViewProperties...> &tgt_field, bool &success,
     Teuchos::FancyOStream &out )
 {
+    // NOTE following line is there to get rid of unused parameter warning.
+    // However, I am not sure why this is a parameter at all since the exact
+    // answer is computed via brute force search of the nearest neighbor and
+    // applying the field function space onto the coordinates of that nearest
+    // neighbor.
+    std::ignore = tgt_field;
+
```
sslattery: Once we change the point cloud operator interface to use multi-rank views we will actually use `tgt_field`. The function copies the output of the generator which is multi-rank to a single rank field view to satisfy the current operator interface.

sslattery: I'm ok with leaving this comment for now so we know to change in the future.

PR #366, opened 2018-02-02
Add Boost.Geometry adapters

dalg24:


packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+TEUCHOS_UNIT_TEST( BoostGeometryAdapters, distance )
+{
+    dtk::Point a = {{0., 0., 0.}};
+    dtk::Point b = {{0., 1., 0.}};
+    TEST_EQUALITY( dtk::distance( a, b ), 1. );
+    TEST_EQUALITY( bg::distance( a, b ), 1. );
+
+    std::tie( a, b ) = std::make_pair<dtk::Point, dtk::Point>( {{0., 0., 0.}},
+                                                               {{1., 1., 1.}} );
```
Rombur: Don't you just need to update `b` not `a` and `b`?

packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,249 @@
...
+    TEST_EQUALITY( dtk::distance( a, b ), std::sqrt( 3. ) );
```
Rombur: I am surprised this work with the round-off...

packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,249 @@
...
+    dtk::Point invalid_point = {{infty, 1.41, 3.14}};
+    TEST_ASSERT( !dtk::isValid( invalid_point ) );
+    TEST_ASSERT( !bg::is_valid( invalid_point, message ) );
+    TEST_EQUALITY( message,
+                   "Geometry has point(s) with invalid coordinate(s)" );
+
+    // Also Boost.Geometry has a is_empty() algorithm but it has a different
+    // meaning, it checks whether a geometry is an empty set and always
+    // return
+    // false for a point or a box.
```
Rombur: can you move the comment on the same line as `return`

dalg24: Will do.  Auto indent didn't do its job very well :)

packages/Search/test/tstBoostGeometryAdapters.cpp
dalg24: Well DTK distance does call `std::sqrt()` so if anything I was more expecting it to fail with Boost.
I will move the NOTE about testing for floating point equality here.  I think it is fine to keep it the way for now.

packages/Search/test/tstBoostGeometryAdapters.cpp
dalg24: I have been moving blocks around that's why I was redefining both A and B.  I think it is pretty clear what I am testing for this way.  Only times I didn't reassign was when name was descriptive like `unit_box` or `origin`.

Rombur: OK it's just funny that you create a pair that you unpack right away 🙂

packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+TEUCHOS_UNIT_TEST( BoostGeometryAdapters, expand )
+{
+    using dtk::equals;
+    dtk::Box box;
+    // NOTE even though not considered valid, default constructed DTK box can be
+    // expanded using Boost.Geometry algorithm.
+    TEST_ASSERT( !bg::is_valid( box ) );
+    bg::expand( box, dtk::Point{{0., 0., 0.}} );
+    dtk::expand( box, {{1., 1., 1.}} );
```
sslattery: so these `expand` functions expand the box to encapsulate the point?

packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,248 @@
...
+    TEST_ASSERT(
+        equals( bg::return_centroid<dtk::Point>( unit_box ), {{.5, .5, .5}} ) );
+    TEST_ASSERT( equals( dtkReturnCentroid( unit_box ), {{.5, .5, .5}} ) );
+
+    // NOTE DTK does not have an overload of centroid() for Point at the
+    // moment.
+    dtk::Point a_point = {{1., 2., 3.}};
+    TEST_ASSERT(
+        equals( bg::return_centroid<dtk::Point>( a_point ), a_point ) );
+    // TEST_ASSERT( equals( dtk::centroid(
```
sslattery: Why is this commented out?

packages/Search/test/tstBoostGeometryAdapters.cpp
```diff
@@ -0,0 +1,248 @@
...
+    //         dtk::centroid( p, c );
+    //         return c;
+    //     }( a_point ),
+    //     a_point ) ) );
+}
+
+TEUCHOS_UNIT_TEST( BoostGeometryAdapters, is_valid )
+{
+    // NOTE "empty" box is considered as valid in DataTransferKit but it is
+    // not according to Boost.Geometry
```
sslattery: What's the difference in how they determine validity?

packages/Search/test/tstBoostGeometryAdapters.cpp
dalg24: Correct.

packages/Search/test/tstBoostGeometryAdapters.cpp
dalg24: Look at the note.  DTK does not have an overload `Point centroid( Point const &)`.

packages/Search/test/tstBoostGeometryAdapters.cpp
dalg24: Look at the messages below.  Boost.Geometry claims to follow the [OGC](http://www.opengeospatial.org) standards on geometry.

PR #367, opened 2018-02-09
Add Boost.Range adapters for Kokkos::View

dalg24:


packages/Search/test/tstBoostRangeAdapters.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include <boost/range/algorithm.hpp> // reverse_copy, replace_if, count
+#include <boost/range/numeric.hpp>   // accumulate
+
+#include "DTK_BoostRangeAdapters.hpp"
+
+#include <sstream>
+
+TEUCHOS_UNIT_TEST( BoostGeometryAdapters, Range )
+{
+    Kokkos::View<int[4], Kokkos::HostSpace> w( "w" );
```
Rombur: Can you add a test where you work with something else than a POD?

dalg24: I amended changes.   Computing an approximation to pi :)

packages/Search/test/tstBoostRangeAdapters.cpp
```diff
@@ -0,0 +1,74 @@
...
+    double const radius = 1.;
+    // 4/3 pi 1^3 / 2^3
+    double const pi = 6. *
+                      static_cast<double>( boost::count_if(
+                          cloud,
+                          [origin, radius]( Point point ) {
+                              return ( distance( point, origin ) <= radius );
+                          } ) ) /
+                      static_cast<double>( n );
+    std::cout << pi << "\n";
```
Rombur: Do you really want to print this on the screen?

dalg24: Nah I forgot to remove it :)

PR #370, opened 2018-02-20
Add not implemented exception

Rombur:
I will need this exception quite a lot in `Interpolation` because I don't plan to support high order finite element right away.

### DIFF
--- a/packages/Utils/test/tstDBC.cpp
+++ b/packages/Utils/test/tstDBC.cpp
@@ -305,5 +305,30 @@ TEUCHOS_UNIT_TEST( DataTransferKitException, errorcode_test_2 )
 }
 
 //---------------------------------------------------------------------------//
+// Check that we can catch a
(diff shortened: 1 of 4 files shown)


packages/Utils/test/tstDBC.cpp
```diff
@@ -304,6 +304,31 @@ TEUCHOS_UNIT_TEST( DataTransferKitException, errorcode_test_2 )
...
+    }
+    catch ( const DataTransferKit::DataTransferKitNotImplementedException
+                &assertion )
+    {
+        message = std::string( assertion.what() );
+    }
+    catch ( ... )
+    {
+        TEST_ASSERT( 0 );
+    }
```
dalg24: Use `TEST_THROW()` macro from `Teuchos` unit testing framework.

dalg24: On second thoughts, this is good enough for now.  We can change whenever we revisit exceptions in DTK.

PR #377, opened 2018-03-26
Update instructions to build on CADES SHPC Condo and allow config with Boost 1.61

dalg24:


packages/Meshfree/test/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ TRIBITS_ADD_EXECUTABLE_AND_TEST(
   FAIL_REGULAR_EXPRESSION "data race;leak;runtime error"
   )
 
-IF (HAVE_DTK_BOOST)
+IF (HAVE_DTK_BOOST AND ((BOOST_VERSION VERSION_EQUAL 1.62.0) OR (BOOST_VERSION VERSION_GREATER 1.62.0)))
```
sslattery: So this test only works with version 1.62 or greater?

scripts/cades_shpc_condo_cmake
```diff
@@ -11,12 +9,12 @@ ARGS=(
     -D CMAKE_BUILD_TYPE=Debug
     -D BUILD_SHARED_LIBS=ON
     ### COMPILERS AND FLAGS ###
-    -D CMAKE_CXX_FLAGS="-g -arch=${GPU_ARCH} -lineinfo \
```
sslattery: so now extended lambda only gets passed to cxx11 flags? What exactly does extended lambda do again?

dalg24: See  #353

packages/Meshfree/test/CMakeLists.txt
dalg24: See #339

scripts/cades_shpc_condo_cmake
dalg24: https://devblogs.nvidia.com/new-compiler-features-cuda-8/

PR #381, opened 2018-04-17
Boost RTree helpers

dalg24:


packages/Search/test/DTK_BoostRTreeHelpers.hpp
```diff
@@ -0,0 +1,115 @@
+#ifndef DTK_BOOST_RTREE_HELPERS_HPP
```
Rombur: copyright is missing

packages/Search/test/DTK_BoostRTreeHelpers.hpp
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return std::make_pair( v.value(), v.index() );
+    }
+};
+
+template <typename Indexable>
+using RTree =
+    boost::geometry::index::rtree<std::pair<Indexable, int>, Parameter>;
+
+// NOTE: The template parameter Indexable is used as an alias here
```
Rombur: Why do you do that? Can I use a different `Indexable`?

packages/Search/test/DTK_BoostRTreeHelpers.hpp
```diff
@@ -0,0 +1,115 @@
...
+    inline bool operator()( Value const &val ) const { return _pred( val ); }
+    static UnaryPredicate makeAlwaysFalse()
+    {
+        return UnaryPredicate( []( Value const & ) { return false; } );
+    }
+    Function _pred;
+};
+
+template <typename Value>
+static auto translate(
```
Rombur: So these two functions transform a DTK query in a  boost query?

packages/Search/test/DTK_BoostRTreeHelpers.hpp
```diff
@@ -0,0 +1,115 @@
...
+translate( DataTransferKit::Details::Nearest<Geometry> const &query )
+    -> decltype( boost::geometry::index::nearest( Geometry(), 0 ) )
+{
+    auto const geometry = query._geometry;
+    auto const k = query._k;
+    return boost::geometry::index::nearest( geometry, k );
+}
+
+template <typename Indexable, typename InputView,
+          typename OutputView = Kokkos::View<int *, Kokkos::HostSpace>>
```
Rombur: Does that work with `OutputView` on the GPU and UVM?

packages/Search/test/DTK_BoostRTreeHelpers.hpp
dalg24: No you cannot.  I was trying to avoid writing `typename View::value_type` more than once :/

packages/Search/test/DTK_BoostRTreeHelpers.hpp
dalg24: Correct.

packages/Search/test/DTK_BoostRTreeHelpers.hpp
dalg24: `InputView` memory space must be accessible from host (static assertion in `DTK_BoostRangeAdapters.hpp`)

I did not anticipated that `OutputView` would be a view on the device and I fill the indices directly from the host in `performQueries()`.  That does not "have to be" this way.

I guess it is me again abusing templates and using `OutputView` as an alias so that I can use it both when I spell out the return type of the function and when I declare the view later on.

packages/Search/test/DTK_BoostRTreeHelpers.hpp
dalg24: Done

packages/Search/test/DTK_BoostRTreeHelpers.hpp
dalg24: @Rombur want me to change anything there?

Rombur: No if it doesn't have to be on the host space then that's fine

Issue #379, opened 2018-03-29
Clang-tidy


Rombur:
Starting from CMake 3.6, CMake has built-in support for clang-tidy. This means that clang-tidy can be used by simply adding sth like: `-DCMAKE_CXX_CLANG_TIDY="/usr/bin/clang-tidy-5.0;-checks=performance-*,mpi-*,portability-*,readability-*"` during configuration. Right now, we are using CMake 3.5 in Docker so we will need to update CMake. The list of acceptable flags for clang-tidy is [here](http://clang.llvm.org/extra/clang-tidy/#using-clang-tidy).

dalg24:
I wonder if we can use `.clang-tidy` configuration file with this.

masterleinad:
`DataTransferKit` is going to be archived.

PR #382, opened 2018-04-17
[WIP] Update base image used on Jenkins CI

dalg24:


docker/Dockerfile_stack
```diff
@@ -245,6 +246,8 @@ RUN echo '#!/usr/bin/env bash' > /usr/local/bin/mpiexec && \
 
 # setup vim
 COPY .vimrc /root/.vimrc
+# FIXME: workaround for CMake (< 3.11) to detect Boost.Python (>= 1.67) with version suffix
+RUN ln -s ${BOOST_DIR}/lib/libboost_python27.so ${BOOST_DIR}/lib/libboost_python.so
```
dalg24: Note that an alternative to this would be to install the latest cmake (3.11.1)
I have not verified whether it would let us work with clang-tidy then (c.f. #379)

docker/Dockerfile_stack
```diff
@@ -45,8 +44,10 @@ RUN apt-get update && apt-get install -y \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN pip install --upgrade --no-cache-dir pip && \
-    pip install --no-cache-dir sphinx==1.4 sphinx_rtd_theme breathe==4.6
+RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+    python get-pip.py && \
+    pip install --upgrade --no-cache-dir pip && \
+    pip install --no-cache-dir sphinx==1.4 sphinx_rtd_theme breathe==4.7
```
dalg24: Now I realize that upgrading breathe to 4.7 is not necessarily ok.  I would have to update the base image for the tests that run on Travis as well.

Rombur:
Two things:
 1. should we use cmake 3.11 to avoid complains from a cmake that it doesn't know the version of boost that we are using
 2. clang 6.0 has been released a month ago so we may want to update that too

packages/Interface/src/DTK_CellList.hpp
```diff
@@ -15,13 +15,12 @@
 //---------------------------------------------------------------------------//
 
 #ifndef DTK_CELLLIST_HPP
-#define DTK_CELLLIST_HPP
+#    define DTK_CELLLIST_HPP
```
Rombur: Actually this looks really strange let's not do that

.clang-format
```diff
@@ -7,10 +7,13 @@ Language:        Cpp
 #   - IndentWidth: 2 -> 4
 #   - SpacesInParentheses: false -> true
 #   - BreakConstructorInitializersBeforeComma: false -> true
+#   - IndentPPDirectives: None -> AfterHash
+#   - IncludeBlocks: Preserve -> Regroup
 AccessModifierOffset: -2
-AlignAfterOpenBracket: true
```
dalg24: @Rombur I generated new LLVM style config file and applied the changes you reported.
It looks like `AlignAfterOpenBracket` and `AlignEscapedNewlinesLeft` differ.  Do you know if it is something you did and forgot to write down or an actual change in the LLVM style?

packages/Interface/src/DTK_CellList.hpp
dalg24: OK so no change `IndentPPDirectives`

.clang-format
Rombur: They change their style. I made sure at the time to record the changes I made

.clang-format
```diff
@@ -7,10 +7,13 @@ Language:        Cpp
 #   - IndentWidth: 2 -> 4
 #   - SpacesInParentheses: false -> true
 #   - BreakConstructorInitializersBeforeComma: false -> true
+#   - IndentPPDirectives: None -> AfterHash
+#   - IncludeBlocks: Preserve -> Regroup
```
dalg24: Consider using `BasedOnStyle: LLVM` and only specify what we actually change

docker/Dockerfile_stack
dalg24: Still needed the workaround with cmake-3.11
I am wondering if it is because ycm try to find boost with the component python instead of pythonX.Y

packages/Meshfree/test/PointCloudProblemGenerator/ExodusProblemGenerator_def.hpp
```diff
@@ -12,20 +12,15 @@
...
 #include <Teuchos_Array.hpp>
-
 #include <Tpetra_Distributor.hpp>
-
-#include <DTK_DBC.hpp>
-
 #include <fstream>
 #include <iostream>
 #include <limits>
+#include <netcdf.h>
```
dalg24: @Rombur this make me like `IncludeBlocksStyle` `Preserve` better than `Regroup`

packages/Search/test/tstLinearBVH.cpp
```diff
@@ -9,19 +9,16 @@
...
-#include <DTK_LinearBVH.hpp>
+#include "Search_UnitTestHelpers.hpp"
 
+#include <DTK_LinearBVH.hpp>
 #include <Teuchos_UnitTestHarness.hpp>
-
-#include <boost/geometry.hpp>
-
 #include <algorithm>
+#include <boost/geometry.hpp>
```
dalg24: Same here, I think `IBS_Preserve` would work better

packages/Utils/src/DTK_DBC.hpp
```diff
@@ -63,7 +63,9 @@ class DataTransferKitException : public std::logic_error
     }
 
     //! Destructor.
-    virtual ~DataTransferKitException() throw() { /* ... */}
+    virtual ~DataTransferKitException() throw()
+    { /* ... */
+    }
```
dalg24: How about that one :/

docker/README.md
```diff
@@ -25,3 +25,16 @@ Clear things up about install prefix of LLVM OpenMP
...
 - Enable auto formatting for Vim within the dev container
+
+## 18.04.0 (April 18th, 2018)
+### TPLs update
+- Updated Boost from 1.65.1 to 1.67.0
+- Updated Trilinos from 89b8c7f (12.12.1) to a27130b (HEAD on master)
+- Updated Kokkos Profiling and Debugging Tools to version tagged as release 2.5.00 on GitHub
+- Updated HDF5 from 1.10.1 to 1.10.2
+- Updated NetCDF from 4.4.1.1 to 4.6.1
+## Change
```
dalg24: missing one `#` here

dalg24:
Ping on this

.clang-format
```diff
@@ -20,27 +21,65 @@ AllowShortFunctionsOnASingleLine: All
...
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: true
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: true
+  AfterStruct:     true
+  AfterUnion:      true
+  AfterExternBlock: false
```
Rombur: Is it worth to define a custom style, for this single options that affects only one file?

docker/README.md
```diff
@@ -25,3 +25,16 @@ Clear things up about install prefix of LLVM OpenMP
...
+- Upgrade CMake to v3.11.1
+- Upgrade all LLVM packages to 6.0.0
+COMMENT: Install OpenMP runtime using system package manager instead of building from source
```
Rombur: So now clang is using `libgomp` (GCC OpenMP) instead of `libomp` (Intel OpenMP)?

packages/Search/src/details/DTK_DetailsBox.hpp
```diff
@@ -8,8 +8,8 @@
  *                                                                          *
  * SPDX-License-Identifier: BSD-3-Clause                                    *
  ****************************************************************************/
-#ifndef DTK_Box_HPP
-#define DTK_Box_HPP
+#ifndef DTK_BOX_HPP
```
Rombur: Nice did you change it or did clang-format do it?

.clang-format
dalg24: I actually had to adjust `AfterEnum` as well.  I have no opinion on this.

docker/README.md
dalg24: I forgot to update this comment. OpenMP now comes from the LLVM+Clang binaries.

packages/Search/src/details/DTK_DetailsBox.hpp
dalg24: I did it myself unfortunately :/

docker/Dockerfile_stack
```diff
@@ -159,10 +157,10 @@ RUN export HDF5_VERSION=1.10.1 && \
     rm -rf ${HDF5_BUILD_DIR} && \
     rm -rf ${HDF5_SOURCE_DIR}
 
-ENV HDF5_DIR=/opt/hdf5/1.10.1
+ENV HDF5_DIR=/opt/hdf5/1.10.2
```
sslattery: is this the latest hdf5?

packages/Discretization/test/tstPointSearch.cpp
```diff
@@ -16,9 +16,8 @@
 #include <Teuchos_UnitTestHarness.hpp>
 
 template <typename DeviceType>
-Kokkos::View<double * [3], DeviceType>
-getPointsCoord3D( Teuchos::RCP<const Teuchos::Comm<int>> comm )
-{
+Kokkos::View<double *[3], DeviceType>
+getPointsCoord3D( Teuchos::RCP<const Teuchos::Comm<int>> comm ) {
```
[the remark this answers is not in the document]
sslattery: yeah don't love the brackets at the end like that

packages/Interface/src/DTK_View.hpp
```diff
@@ -72,7 +72,7 @@ class View
     View( KokkosViewType kokkos_view,
           typename std::enable_if<
               Kokkos::is_view<KokkosViewType>::value ||
-                  Kokkos::Experimental::is_dyn_rank_view<KokkosViewType>::value,
+                  Kokkos::is_dyn_rank_view<KokkosViewType>::value,
```
sslattery: nice - glad this isn't experimental anymore

packages/Utils/src/DTK_DBC.hpp
sslattery: yeah that sucks but it probably always wants to separate the '{}'

docker/Dockerfile_stack
dalg24: yes

packages/Utils/src/DTK_DBC.hpp
dalg24: Actually if you look at `DTK_DBC.hpp`, you will see that the two constructors for `DataTransferKitException` do this as well. This destructor should probably be `= default;` anyway.

PR #385, opened 2018-05-01
Miscellaneous changes to Search package including moving predicates out of Details namespace

dalg24:
I had these local changes sitting on a work in progress branch on fetnat01.  It felt like I should push it to github just in case maintenance goes wrong this afternoon.

These are miscellaneous changes to the Search package.  One worth mentioning is I am moving the predicates out of the `Details::` namespace.  This needed to happen because it is part of the interface.

I figured I might as well open a PR.


dalg24:
Note that I experimented with renaming `BVH` to `BoundingVolumeHierarchy` while providing an alias so that changes don't need to be propagated for now.  I can drop this if it feels too silly.

PR #388, opened 2018-05-02
Run on Summitdev

hydroo:
The magic is 
```
-D MPI_EXEC="jsrun"
-D MPI_EXEC_NUMPROCS_FLAG="-n"
-D MPI_EXEC_POST_NUMPROCS_FLAGS="-a1;-c5;-bpacked:5"
```
- semicolons in MPI_EXEC_POST_NUMPROCS_FLAGS instead of spaces!
- `-n` means number of resource sets
- `-a1` means one task/mpirank per resource set
- `-c5` means use 5 cores per resource set
- `bpacked` means threads per task/mpirank.
- `-g1`means one GPU per resource set

Since in cmake max procs for MPI is set to 4, I've set 5 threads per rank maximum to not create more threads then there are cores.
Technically -c and -bpacked are supposed to be total cores (20) divided by number of ranks.

If you want more threads, you'd either make -c and -bpacked adaptable or live with using hyperthreading / multiple threads per core.

### DIFF
--- a/scripts/olcf_summitdev_cmake
+++ b/scripts/olcf_summitdev_cmake
@@ -5,11 +5,11 @@ ARGS=(
     -D BUILD_SHARED_LIBS=ON
     ### COMPILERS AND FLAGS ###
     -D CMAKE_CXX_FLAGS="-g -arch=sm_60 -lineinfo \
-        -std=c++11 --expt-extended-lambda \
         -Xcudafe --diag_suppress=conversion_function_not_usable \
         -Xcudafe --diag_suppress=cc_clobber_ignored \
         -Xcudafe --diag_suppress=code_is_unreachable"
     -D DataTransferKit_CXX_FLAGS="-Wall -Wno-shadow -Wpedantic"
+    -D Trilinos_CXX11_FLAGS="-std=c++11 --expt-extended-lambda"
     ### TPLs ###
     -D TPL_ENABLE_MPI=ON
     -D TPL_ENABLE_BLAS=ON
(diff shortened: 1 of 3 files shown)


dalg24:
ok to test

@hydroo Were you actually able to run all the tests on summitdev using ctest after you configured with these flags?

hydroo:
Jsrun is awfully slow. And the graphicscard setup seems wrong (I'll fix that). Other than that it looks promissing. Will keep you updated.

dalg24:
You either want to disable the CUDA ETI or pass `-g1` to `jsrun`

hydroo:
I sent the ctest logs to @dalg24. Looks OK (Except `DataTransferKitDiscretization_PointSearch_MPI_4` is segfaulting on Peak).

Should I mention somewhere that these instruction work on Summit?
I.e. rename things to just "summit" or "summitdev_and_summit"?

scripts/olcf_summitdev_cmake
```diff
@@ -1,15 +1,23 @@
...
     -D CMAKE_CXX_FLAGS="-g -arch=sm_60 -lineinfo \
-        -std=c++11 --expt-extended-lambda \
         -Xcudafe --diag_suppress=conversion_function_not_usable \
         -Xcudafe --diag_suppress=cc_clobber_ignored \
         -Xcudafe --diag_suppress=code_is_unreachable"
     -D DataTransferKit_CXX_FLAGS="-Wall -Wno-shadow -Wpedantic"
+    -D Trilinos_CXX11_FLAGS="-std=c++11 --expt-extended-lambda"
+    -D MPI_EXEC="jsrun"
+    -D MPI_EXEC_NUMPROCS_FLAG="-n"
+    -D MPI_EXEC_POST_NUMPROCS_FLAGS="-a1;-c5;-bpacked:5;-g1"
```
dalg24: We might want to use a variable for the number of threads.

PR #399, opened 2018-06-06
Check precondition on source view in NearestNeighborOperator::apply()

dalg24:


packages/Meshfree/src/DTK_DetailsNearestNeighborOperatorImpl.hpp
```diff
@@ -188,6 +188,20 @@ struct NearestNeighborOperatorImpl
...
+    // TODO move this to comm helpers headers together with sendAcrossNetwork()
+    // and fetch()
+    static bool allReduceBooleanLogicalAnd( Teuchos::Comm<int> const &comm,
+                                            bool val )
+    {
+        int res = 0;
+        Teuchos::reduceAll( comm, Teuchos::REDUCE_SUM, val ? 0 : 1,
+                            Teuchos::ptrFromRef( res ) );
+        return res;
+    }
```
dalg24: Haven't bothered writing standalone test for this one in `tstDetailsCommunicationHelpers.cpp` :/

dalg24:
After talking to @Rombur, I realize that I shouldn't be doing the MPI Allreduce to circumvent the limitations from the Teuchos unit test framework (MPI abort not called when exception is thrown only on one process which may cause a deadlock)

I decided to revert the commits that introduce `allReduceBooleanLogicalAnd()` rather than dropping them so that we still have it around if we later decide we'd like to have such a functionality after all.

PR #403, opened 2018-06-19
Add CUDA 9.2 pkg version to setup script for nvidia-docker

dalg24:


dalg24:
@sslattery  I updated the documentation.  Let me know if that works with you.

sslattery:
Here is my build output:
```

sslattery:
Looks I got some memory errors in CUDA. The failing code also had some build warnings.

PR #404, opened 2018-06-20
Silent CUDA9 warnings with defaulted function marked as KOKKOS_FUNCTION

dalg24:
warning: __device__ annotation on a defaulted function("XXXXX") is ignored
warning: __host__ annotation on a defaulted function("XXXXX") is ignored

c.f. kokkos/kokkos#1470


dalg24:
`-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored` is only valid from CUDA 9.0 so it cannot just be added to our CUDA configure script :/

dalg24:
We probably want to conditionally add the flag but this cannot be done directly in the configuration script.  I can open a separate issue to track this problem if you like.  Let me know.

PR #410, opened 2018-07-12
Sort queries to reduce data divergence

dalg24:
An attempt to take advantage of the fact that if objects are nearby, chances are that they will yield almost the same traversals.

### DIFF
--- a/packages/Search/src/details/DTK_DetailsBatchedQueries.hpp
+++ b/packages/Search/src/details/DTK_DetailsBatchedQueries.hpp
@@ -1,0 +1,188 @@
+/****************************************************************************
+ * Copyright (c) 2012-2018 by the DataTransferKit authors                   *
+ * All rights reserved.                                                     *
+ *                                                                          *
+ * This file is part of the DataTransferKit library. DataTransferKit is     *
+ * distributed under a BSD 3-clause license. For the licensing terms see    *
+ * the LICENSE file in the top-level directory.                             *
+ *                                                                          *
+ * SPDX-License-Identifier: BSD-3-Clause                                    *
+ ****************************************************************************/
+
+#ifndef DTK_DETAILS_BATCHED_QUERIES_HPP
+#define DTK_DETAILS_BATCHED_QUERIES_HPP
+
+#include <DTK_Box.hpp>
+#include <DTK_DetailsAlgorithms.hpp>       // return_centroid
+#include <DTK_DetailsTreeConstruction.hpp> // morton3D
+#include <DTK_DetailsUtils.hpp> // iota, exclusivePrefixSum, lastElement
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Parallel.hpp>
+#include <Kokkos_View.hpp>
+
+#include <tuple>
+
+namespace DataTransferKit
+{
+
+namespace Details
+{
+template <typename DeviceType>
+struct BatchedQueries
+{
+  public:
+    using ExecutionSpace = typename DeviceType::execution_space;
+
+    // BatchedQueries defines functions for sorting queries along the Z-order
+    // space-filling curve in order to minimize data divergence.  The goal is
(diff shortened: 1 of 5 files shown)


dalg24:
cuda (P100)  | e9a2bde  | 75ef59b  | 4f133ce7
------------ | -------- | -------- | --------
construction | 0.319332 | 0.312941 | 0.308305
kNN          | 0.394494 | 0.40688  | 0.382948
radius       | 0.264993 | 0.315028 | 0.348413

openmp       | e9a2bde  | 75ef59b  | 4f133ce7
------------ | -------- | -------- | --------
construction | 0.348495 | 0.348303 | 0.350486
kNN          | 0.366683 | 0.368394 | 0.583663
radius       | 0.380238 | 0.38536  | 0.86287

serial       | e9a2bde  | 75ef59b  | 4f133ce7
------------ | -------- | -------- | --------
construction | 3.56007  | 3.50881  | 3.52007
kNN          | 4.29842  | 4.27737  | 6.61133
radius       | 4.51795  | 4.52966  | 9.11296

cuda (K80)   | e9a2bde  | 75ef59b  | 4f133ce7
------------ | -------- | -------- | --------
construction | 1.21553  | 1.21178  | 1.21017
kNN          | 2.07129  | 2.11973  | 2.16152
radius       | 0.962253 | 1.01692  | 1.69827

dalg24:
Timings above were obtained with `--values=10000000 --queries=1000000`.  Note that as the size of the tree and the number of queries decrease, sorting the queries gets less and less advantageous and it eventually penalize us for small problems.

dalg24:
For the record, I did try to also add a level of indirection when accessing queries instead of actually applying the permutation but I found it was actually a bit slower.

packages/Search/src/details/DTK_DetailsBatchedQueries.hpp
```diff
@@ -0,0 +1,188 @@
...
+    permuteIndices( Kokkos::View<int const *, DeviceType> permute,
+                    Kokkos::View<T const *, DeviceType> indices,
+                    Kokkos::View<int const *, DeviceType> offset,
+                    Kokkos::View<int const *, DeviceType> tmp_offset )
+    {
+        auto const n = permute.extent( 0 );
+
+        DTK_REQUIRE( offset.extent( 0 ) == n + 1 );
+        DTK_REQUIRE( tmp_offset.extent( 0 ) == n + 1 );
+        DTK_REQUIRE( lastElement( offset ) == indices.extent_int( 0 ) );
```
Rombur: Be consistent (at least inside one function), either use `extent` or `extent_int` but don't mix them

packages/Search/test/tstDetailsBatchedQueries.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    TEST_THROW( DataTransferKit::Details::BatchedQueries<
+                    DeviceType>::reversePermutation( permute, offset, indices ),
+                DataTransferKit::DataTransferKitException );
+
+    Kokkos::resize( offset, 1 );
+    TEST_NOTHROW( DataTransferKit::Details::BatchedQueries<
+                  DeviceType>::reversePermutation( permute, offset, indices ) );
+
+    DataTransferKit::Details::BatchedQueries<DeviceType>::permuteOffset(
```
Rombur: Is this a test? If yes either comment or use `TEST_NOTHROW`

dalg24: I think I just forgot to remove that line.

packages/Search/src/details/DTK_DetailsBatchedQueries.hpp
dalg24: Ok, will fix.

dalg24:
OpenMP runs were obtained with
```
OMP_PROC_BIND=spread
OMP_NUM_THREADS=16
OMP_PLACES=threads
```

PR #419, opened 2018-08-08
Configure with TPL Boost installed with system package manager

dalg24:
@sslattery does that resolve the issue you were seeing when trying to configure against Boost from your package management system?


dalg24:
Updated.  Fix was trivial.  Error was we were relying on an optional variable that may or may not have been passed at configure time instead of using the directory path set by TriBITS.

PR #445, opened 2018-09-20
Improve installation documentation when using a tarball

Rombur:
The current instructions to install DTK do not work if we are downloading a tarball. 

I will cherry-pick this commit.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -65,6 +65,15 @@ Another way to achieve this is to pass the ``--recursive`` option to the ``git
 clone`` command which will automatically initialize and update DTKData in the
 DataTransferKit repository.
 
+You can also download the DTKData
+`repository <https://github.com/ORNL-CEES/DTKData>`_ yourself and then, link
+it into DTK directory:
+
+.. code::
+
+  $ cd $DTK_DIR
+  $ ln -s $DTKDATA_DIR DTKData
+
 Building DTK
 ------------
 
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -65,6 +65,15 @@ Another way to achieve this is to pass the ``--recursive`` option to the ``git
...
 DataTransferKit repository.
 
+You can also download the DTKData
+`repository <https://github.com/ORNL-CEES/DTKData>`_ yourself and then, link
+it into DTK directory:
+
+.. code::
+
+  $ cd $DTK_DIR
+  $ ln -s $DTKDATA_DIR DTKData
```
dalg24: Would need to define env variable `DTKDATA_DIR` for that to work.
I realize now that there is the same problem with `DTK_DIR` but the goal would be that people can almost copy/paste from these instructions.

Rombur: Sure but I assumed that people were smart enough to replace the `$DTKDATA_DIR` with the right path.

dalg24: I usually copy and paste without thinking and add `sudo` in front of it if it didn't work.

Rombur: I'll make sure to add `rm -rf /` at the end of the instruction then.

dalg24:
```
root@00f6503f974c:/# rm -rf /
rm: it is dangerous to operate recursively on '/'
rm: use --no-preserve-root to override this failsafe
```

PR #449, opened 2018-09-20
Do not require Boost.MPI and Boost.Serialization in release 3.0

dalg24:
It removes the need for building Boost


dalg24:
> This will have to be cherry-picked

No I did PR onto dtk-3.0 branch

PR #450, opened 2018-09-20
Fortran fixes

dalg24:
Go ahead and piggy back fixes for PGI on this PR


Rombur:
I've removed the `ffree-line-length-none` flag. With this branch pgi works.

Rombur:
I think this can be merged. It will need to be cherry-picked.

PR #458, opened 2018-09-26
Remove C++ API from documentation published on ReadTheDocs

dalg24:


dalg24:
Will need to be merged/cherry-picked into `dtk-3.0` as well.

PR #469, opened 2018-10-02
Performance test

Rombur:
This PR adds a regression test on the performance.This works only for one build (CUDA build) but we can change that in the future. For now, we don't use repetition, the python script `compare.py` needs to be modified to deal with repetitions. We can extend the capabilities later but I would like to see if everything works well before doing so.

### DIFF
--- a/docker/jenkins/build.sh
+++ b/docker/jenkins/build.sh
@@ -50,3 +50,9 @@ curl -s https://codecov.io/bash -o codecov_bash_uploader
 chmod +x codecov_bash_uploader
 ./codecov_bash_uploader -Z -X gcov -f lcov.info
 fi
+# Run performance tests
+if [ "${BUILD_TYPE}" == "gcc54-cuda8" ]
+then
+  ./DataTransferKit/packages/Search/examples/bvh_driver/DataTransferKitSearch_bvh.exe --benchmark_out=test.json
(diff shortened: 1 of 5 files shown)


docker/jenkins/build.sh
```diff
@@ -50,3 +50,9 @@ curl -s https://codecov.io/bash -o codecov_bash_uploader
 chmod +x codecov_bash_uploader
 ./codecov_bash_uploader -Z -X gcov -f lcov.info
 fi
+# Run performance tests
+if [ "${BUILD_TYPE}" == "gcc54-cuda8" ]
+then
+  ./DataTransferKit/packages/Search/examples/bvh_driver/DataTransferKitSearch_bvh.exe --benchmark_out=test.json
```
dalg24: So no `--benchmark_repetitions=N`?

scripts/compare.py
```diff
@@ -0,0 +1,109 @@
...
+import os.path
+import sys
+
+# Read the command line arguments
+argv = sys.argv[1:]
+opts, args = getopt.getopt(argv, 'hc:b:r:n:o:t:', ['commit_hash=',
+                                                   'build_number=', 'input_ref=', 'input_new=',
+                                                   'output_file=', 'tolerance='])
+# Default relative tolerance is 10%
+tol = 0.2
```
dalg24: `tol = 0.1`??

Rombur:
Closing while doing some tests

PR #471, opened 2018-10-02
Performance

### LINKED ISSUE #429: Add performance regression testing

Rombur:
New PR because I cannot reopen #469 because I force-pushed.

close #429


Rombur:
_ping_

dalg24:
I already approved.  Go ahead and merge unless you'd like feedback from @aprokop and @sslattery as well.

PR #463, opened 2018-09-27
Fixes for release 3.0

dalg24:
Pushing #456 and #462 into dtk-3.0

PR #472, opened 2018-10-03
Fix a bug using CUDA 9

Rombur:
With this `PointSearch` passes using cuda 9 however `DataTransferKitHybridTransport_CartesianMesh_MPI_4` always hangs. I think this should be cherry-picked for the release.


sslattery:
Weird - so this came down to having arrays of kokkos views?

Rombur:
> Weird - so this came down to having arrays of kokkos views?

Yes with CUDA 9.0 I get a warning about using `__host__` function (operator[]) but not with CUDA 8.0

dalg24:
Merging.  @Rombur would you please cherry-pick into #463

PR #473, opened 2018-10-03
Update CUDA package version for 9.2 in script to extend docker images

dalg24:


Rombur:
Doesn't work
```
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 cuda-command-line-tools-9-2 : Depends: cuda-cupti-9-2 (>= 9.2.148) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

dalg24:
@Rombur Let's remove the 9.2 entry if it doesn't build.

For the record, we were discussing the possibility of building from `nvidia/cuda` images instead of extending our own image.

Rombur:
You should squash the two commits.

PR #499, opened 2018-11-16
Update README for building on CADES SHPC Condo

dalg24:
Main change is using `KOKKOS_ARCH` instead of adding `-arch` flag

`CMAKE_PREFIX_PATH` is in my home directory but it shouldn't prevent anyone else to configure.
You should be able to point to it if you don;'t want to install Benchmark on your own.

### DIFF
--- a/scripts/cades_shpc_condo_cmake
+++ b/scripts/cades_shpc_condo_cmake
@@ -6,7 +6,7 @@ rm -f  CMakeCache.txt
 rm -rf CMakeFiles/
 EXTRA_ARGS=("$@")
 ARGS=(
-    -D CMAKE_BUILD_TYPE=Debug
+    -D CMAKE_BUILD_TYPE=Release
     -D BUILD_SHARED_LIBS=ON
     ### COMPILERS AND FLAGS ###
     -D CMAKE_CXX_FLAGS="-g -lineinfo \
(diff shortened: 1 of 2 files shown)


sslattery:
`Bikeshed in config script for condo (quote paths to TPLs)` - what is `Bikeshed`?

Rombur:
image

dalg24:
http://bikeshed.com

I can change the commit message if you like.  I meant this is a pointless commit where I decided to quote the paths to the TPLs.

dalg24:
> May also want to escape TRILINOS_DIR.

You mean put double quotes around it?

aprokop:
> You mean put double quotes around it?

Sorry, yes, that's what I meant. In general, **everything** in bash should be escaped.

scripts/cades_shpc_condo_cmake
```diff
@@ -56,4 +56,4 @@ ARGS=(
         -D DataTransferKit_ENABLE_TESTS=ON
         -D DataTransferKit_ENABLE_EXAMPLES=ON
     )
-cmake "${ARGS[@]}" "${EXTRA_ARGS[@]}" $TRILINOS_DIR
+cmake "${ARGS[@]}" "${EXTRA_ARGS[@]}" "${TRILINOS_DIR}"
```
dalg24: Sharing credit for all the bikeshed

PR #500, opened 2018-11-27
Free memory allocated for source and target clouds in Fortran API test

dalg24:

PR #501, opened 2018-11-27
[WIP] Add suppression file for possile leak detected by Valgrind

dalg24:
Not investigating this memory error further because I don't think it's worth our time


dalg24:
CI is not setup properly for dtk-2.0 as in #500 
Please ignore failures

dalg24:
Note that Valgrind detects the exact same memory error in the C API test.  So the current location of the suppression file might not be the most appropriate.

PR #503, opened 2018-12-03
Initialize MPI in all tests to resolve issue with CUDA hooks on OLCF machines

dalg24:


packages/Interface/test/tstFortran_API.f90
```diff
@@ -700,11 +700,11 @@ program main
   use datatransferkit
   use x
   use utc
-#ifdef HAVE_MPI
   use mpi
-#endif
   implicit none
 
+  integer :: ierr
```
dalg24: Note that `HAVE_MPI` was never defined and that the code would not compiler because `ierr` was not declared anywhere.

dalg24:
Performance regression fails for no apparent good reason, preventing the merge.  @Rombur How do you suggest we handle this?

PR #504, opened 2018-12-06
Small optimization for the GPU

Rombur:
This decreases the number of copy which slightly improves performance on GPU.

### DIFF
--- a/packages/Search/src/details/DTK_DetailsNode.hpp
+++ b/packages/Search/src/details/DTK_DetailsNode.hpp
@@ -22,6 +22,13 @@ struct Node
     KOKKOS_INLINE_FUNCTION
     Node() = default;
 
+    KOKKOS_INLINE_FUNCTION
+    Node( const Kokkos::pair<Node *, Node *> &c, const Box &bb )
+        : children( c )
+        , bounding_box( bb )
+    {
+    }
+
     Kokkos::pair<Node *, Node *> children = {nullptr, nullptr};
     Box bounding_box;
 };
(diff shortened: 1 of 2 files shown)


packages/Search/src/details/DTK_DetailsTreeConstruction_def.hpp
```diff
@@ -269,9 +269,9 @@ void TreeConstruction<DeviceType>::initializeLeafNodes(
         DTK_MARK_REGION( "initialize_leaf_nodes" ),
         Kokkos::RangePolicy<ExecutionSpace>( 0, n ), KOKKOS_LAMBDA( int i ) {
             // FIXME: Here also accessing node bounding box data member
```
dalg24: Remove that comment.

packages/Search/src/details/DTK_DetailsNode.hpp
```diff
@@ -22,6 +22,13 @@ struct Node
     KOKKOS_INLINE_FUNCTION
     Node() = default;
 
+    KOKKOS_INLINE_FUNCTION
+    Node( const Kokkos::pair<Node *, Node *> &c, const Box &bb )
```
dalg24: This constructor was needed to allow the list-initialization of the leaf nodes in `TreeConstruction:: initializeLeafNodes()`.

dalg24:
Note that I could not measure any improvement on the GPU (<.5% slower than master) whereas I measure a ~2% slow down on the CPU with 32 threads.  (I had `--values=1000000`).

PR #505, opened 2018-12-13
Add support for checking type of primitives for creating a bounding volume hierarchy

dalg24:


packages/Search/src/details/DTK_DetailsConcepts.hpp
```diff
@@ -0,0 +1,68 @@
...
+    Geometry, Other,
+    std::void_t<decltype(
+        expand( std::declval<Geometry &>(), std::declval<Other const &>() ) )>>
+    : std::true_type
+{
+};
+
+// Checks for existence of a free function that calculates the centroid of an
+// object of type Geometry
+template <typename Geometry, typename Point, typename = void>
```
Rombur: Why is `Point` a template?

dalg24: Might be overkill at the moment.  I had in mind to allow customization.  We'd only be able to access the coordinates through some traits when we compute Morton codes, we don't actually need the type to be our `DTK::Point`.  Also didn't feel like including any other DTK headers.
You want me to change it?

dalg24:
>  can you squash a few commits (fix typo, remove type modification,...)

Okay, will clean history just before the merge.

packages/Search/src/details/DTK_DetailsConcepts.hpp
Rombur: No, that's fine. I don't have a strong opinion on this.

dalg24:
(Rebased on #507 and pushed force)

PR #520, opened 2019-01-15
Update script to setup Clang in the Docker container

dalg24:
Rational:
* there was no default-clause in the if statement that would determine what sanitizer to use
* the thread sanitizer does not work with OpenMP so the output was total garbage (it produced ~200MB of error when running one of our tests!)

I chose to setup the sanitizer all the time.  Let me know if you think that's problematic.

### DIFF
--- a/scripts/docker_clang60_env.sh
+++ b/scripts/docker_clang60_env.sh
@@ -7,15 +7,18 @@ export CXX=mpicxx
 # Add clang OpenMP runtime library to LD_LIBRARY_PATH
 export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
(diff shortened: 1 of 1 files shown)


dalg24:
> It's fine with me. There was no default sanitizer because the default was to have no sanitizer. With the new script, you have no choice and the sanitizer will run no matter what.

How about we do not enable the sanitizer by default but if **any** argument is passed to the script we set up the sanitizer?

Rombur:
> How about we do not enable the sanitizer by default but if any argument is passed to the script we set up the sanitizer?

Sounds good.

dalg24:
@Rombur bruno I reverted the changes to `docker/jenkins/build.sh`

PR #522, opened 2019-01-17
Fix a bug with gcc-8.2

### LINKED ISSUE #519: Interface does not support gcc 8.2
I tried to compiled DTK with gcc 8.2 but I get tens of error from the `Interface` package. There are all similar to the one below
```
DataTransferKit/packages/Interface/test/tstUserApplication.cpp:408:27:
error: passing ‘const DataTransferKit::View<long long int>’ as ‘this’ argument discards
 qualifiers [-fpermissive]        
         global_dof_ids[n] = n + offset;
```
`Interface` is the only package which doesn't work. If I comment all the bugged lines, I can compile everything else and pass the tests.

Rombur:
gcc was choosing the wrong overload. So removed a function that is not
needed.

closes #519

### DIFF
--- a/packages/Interface/src/DTK_View.hpp
+++ b/packages/Interface/src/DTK_View.hpp
@@ -109,16 +109,8 @@ class View
     KOKKOS_INLINE_FUNCTION
     const Scalar *data() const { return _data; }
 
-    // Access an element in the view.
-    KOKKOS_FORCEINLINE_FUNCTION
-    Scalar &operator[]( const size_t i )
-    {
-        assert( _data );
-        assert( i < _size );
-        return _data[i];
-    }
-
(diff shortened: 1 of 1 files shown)


packages/Interface/src/DTK_View.hpp
```diff
@@ -109,15 +109,6 @@ class View
...
-    // Access an element in the view.
-    KOKKOS_FORCEINLINE_FUNCTION
-    Scalar &operator[]( const size_t i )
-    {
-        assert( _data );
-        assert( i < _size );
-        return _data[i];
-    }
-
     // Access a const element in the view.
```
dalg24: I would add a "FIXME const-correctness"

PR #524, opened 2019-01-18
Update Boost dependency

Rombur:
I have removed the dependency to the parts of boost that we don't use and added the parts that we use


Rombur:
Codecov was shortly offline for a database update and that wreak havoc all our scripts.

PR #526, opened 2019-01-21
Make dev/ci image derive from nvidia/cuda

### LINKED ISSUE #480: Make our GPU enabled image derive from nvidia/cuda instead of extending ci/dev base image
c.f. #473 and issues with CUDA versions from 9.2 onwards
c.f. #451 to install Trilinos with CUDA enabled

dalg24:
Rather than installing the CUDA toolchain on top of our base image

Closes #480

### DIFF
--- a/docker/nvidia/setup_nvidia_docker_compose.py
+++ b/docker/nvidia/setup_nvidia_docker_compose.py
@@ -120,19 +120,7 @@ for service in extend_services:
     config['services'][service]['devices'] = devices
     config['services'][service]['volumes'] = volumes
     config['services'][service]['environment'] = ['GPU_ARCH=' + gpu_arch]
-    config['services'][service]['image'] = service + '_cuda_toolchain'
-    dockerfile = 'Dockerfile_' + service + '_cuda_toochain_generated'
-    config['services'][service]['build'] = {
-        'context': '.',
-        'dockerfile': dockerfile,
-        'args': [
-            'CUDA_VERSION=' + cuda_version,
-            'CUDA_PKG_VERSION=' + cuda_pkg_version,
-        ],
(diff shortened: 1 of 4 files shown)


dalg24:
Dammit Andrey.  Will rollback the update of Trilinos

dalg24:
PS this is ready for review.  I will **not** rebase on #528 and build new images

docker/nvidia/setup_nvidia_docker_compose.py
```diff
@@ -120,19 +120,7 @@ def get(self, endpoint):
...
     config['services'][service]['volumes'] = volumes
     config['services'][service]['environment'] = ['GPU_ARCH=' + gpu_arch]
-    config['services'][service]['image'] = service + '_cuda_toolchain'
-    dockerfile = 'Dockerfile_' + service + '_cuda_toochain_generated'
-    config['services'][service]['build'] = {
-        'context': '.',
-        'dockerfile': dockerfile,
-        'args': [
-            'CUDA_VERSION=' + cuda_version,
-            'CUDA_PKG_VERSION=' + cuda_pkg_version,
```
Rombur: I think that `cuda_pkg_version` is not used anymore. So we can remove all the code associated with it.

dalg24:
> You can tag the new images and then merge.

I have already tagged them as latest and pushed to Docker Hub.  This is ready to merge.

docker/nvidia/setup_nvidia_docker_compose.py
dalg24: You are correct.  I will address that in another PR when we move towards nvidia-docker 2.0 if that's OK with you.

PR #530, opened 2019-01-30
Remove explicit instantiation in Search package

dalg24:
Also remove the `dtk_search_details` library and "merging it" into `dtk_search`


packages/Search/src/details/DTK_DetailsTreeConstruction_decl.hpp
```diff
@@ -424,6 +429,78 @@ Node *TreeConstruction<DeviceType>::generateHierarchy(
...
+    auto const last = first + leaf_nodes.extent( 0 );
+    Node *root = internal_nodes.data();
+    Kokkos::parallel_for(
+        DTK_MARK_REGION( "calculate_bounding_boxes" ),
+        Kokkos::RangePolicy<ExecutionSpace>( first, last ),
+        CalculateInternalNodesBoundingVolumesFunctor<DeviceType>( root, parents,
+                                                                  first ) );
+    Kokkos::fence();
+}
+
```
Rombur: Only four commits to move the code from `_def` to `_decl`, I am disappointed.

dalg24:
> I think it looks good but let's see if the code times out again. Can you check that [this](https://github.com/ORNL-CEES/DataTransferKit/pull/530/commits/e04de97b8a0ff957e8c14ced47f6618d1ef75d33#diff-fdca7c3ce7ef2b0973147686c4370238R18) is still necessary. This looks very strange since you would expect the language to be CUDA and not CXX

```
CMake Error: CMake can not determine linker language for target: dtk_search
```

Rombur:
@dalg24 Thanks for trying

PR #563, opened 2019-12-05
Update arborx

Rombur:
Update to use the current master of arborx and add a couple fixes that are necessary to build DTK

### DIFF
--- a/packages/Discretization/src/DTK_PointSearch_def.hpp
+++ b/packages/Discretization/src/DTK_PointSearch_def.hpp
@@ -86,8 +86,10 @@ void buildTopo( Kokkos::View<int *, DeviceType> imported_cell_indices,
 }
 
 template <typename ViewType>
-void sendDataAcrossNetwork( ArborX::Details::Distributor const &distributor,
-                            std::pair<ViewType, ViewType> data )
+void sendDataAcrossNetwork(
+    ArborX::Details::Distributor<typename ViewType::device_type> const
+        &distributor,
+    std::pair<ViewType, ViewType> data )
 {
     ArborX::Details::DistributedSearchTreeImpl<
         typename ViewType::device_type>::sendAcrossNetwork( distributor,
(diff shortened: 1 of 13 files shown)


packages/Discretization/src/DTK_PointSearch_def.hpp
```diff
@@ -96,8 +98,9 @@ void sendDataAcrossNetwork( ArborX::Details::Distributor const &distributor,
 }
 
 template <typename T, typename... Targs>
-void sendDataAcrossNetwork( ArborX::Details::Distributor const &distributor,
-                            std::pair<T, T> d, Targs... data )
+void sendDataAcrossNetwork(
+    ArborX::Details::Distributor<typename T::device_type> const &distributor,
+    std::pair<T, T> d, Targs... data )
```
dalg24: Did you mean to forward the pack?

packages/Discretization/src/DTK_PointSearch_def.hpp
```diff
@@ -387,14 +391,16 @@ PointSearch<DeviceType>::performDistributedSearch(
...
                           Kokkos::RangePolicy<ExecutionSpace>( 0, n_points ),
                           KOKKOS_LAMBDA( int i ) {
-                              queries( i ) = ArborX::within(
-                                  {{points_coord( i, 0 ), points_coord( i, 1 ),
-                                    points_coord( i, 2 )}},
-                                  0. );
+                              queries( i ) = ArborX::intersects( ArborX::Sphere{
+                                  {points_coord( i, 0 ), points_coord( i, 1 ),
+                                   points_coord( i, 2 )},
+                                  0.} );
```
dalg24: Intersect with a point would be better.

Rombur:
Both are valid comment and they are actually fixed in my branch that uses the `callback` mechanic. This PR just update ArborX, I would prefer that we wait for the `callback` PR to fix these comments.

Rombur:
retest this please

Rombur:
I think this is ready to be merged. The fixes Damien wants will be done in a later PR where we use the new callback mechanism.

PR #564, opened 2019-12-06
Avoid some compiler warnings

masterleinad:


dalg24:
Add to whitelist

packages/Benchmarks/HybridTransport/src/DTK_Benchmark_MonteCarloMesh.cpp
```diff
@@ -77,12 +77,11 @@ void MonteCarloMesh::partition(
     const std::vector<double> &z_bnd_mesh )
 {
     // Determine how many I, J, and K blocks there will be.
-    int comm_size = comm->getSize();
     int num_i_blocks = x_bnd_mesh.size() - 1;
     int num_j_blocks = y_bnd_mesh.size() - 1;
     int num_k_blocks = z_bnd_mesh.size() - 1;
     int num_blocks = num_i_blocks * num_j_blocks * num_k_blocks;
-    DTK_REQUIRE( num_sets * num_blocks == comm_size );
+    DTK_REQUIRE( num_sets * num_blocks == comm->getSize() );
```
dalg24: What kind of warning was emitted?  Was it about unused variable when the debug assertion macro is disabled?

masterleinad: Yes, exactly.

dalg24: The fix propose is fine by me but so you know, DTK has a [`DTK_REMEMBER()`](https://github.com/ORNL-CEES/DataTransferKit/blob/edfa050cd46e2274ab0a0b7558caca0079c2e4ca/packages/Utils/src/DTK_DBC.hpp#L128-L129) macro that can be used in that case.

dalg24:
Retest this please

PR #566, opened 2019-12-18
Make Trilinos a TPL using Kokkos 3.0

masterleinad:
Based on top of #451, these are the changes required to build with Trilinos with `Kokkos` 3.0.

~~We still need to make sure everything works without using deprecated code in `Kokkos`.~~


masterleinad:
Is this good enough for the milestone? I can confirm that this works for me locally but since it relies on `Kokkos-3.0` it doesn't pass `CI` at all. I guess it only makes sense to clean this up once we have the release and also `Trilinos` is updated.

Rombur:
So what do we need to get this merged? Just update the CI to Kokkos 3.0 and Trilinos master?

masterleinad:
> So what do we need to get this merged? Just update the CI to Kokkos 3.0 and Trilinos master?
Admittedlty
Yes, that is what I tested back then. Admittedly, I am not quite sure how to modify CI accordingly.

Rombur:
Can you post your configure line.

masterleinad:
Apart from
```
DataTransferKit_CXX11_FLAGS=-std=c++14
```
I don't need anything special to compile using a recent `ArborX` version.

dalg24:
Close?

Rombur:
Yes, the commits are in master

PR #569, opened 2020-03-16
Update Docker and CI

Rombur:
This is not ready for review. I am just trying to see what I need to do to update the CI


Rombur:
This superseeds #451 and #566 It's not perfect but I think that we should merge this and improve on it later. We only test one configuration (serial + OpenMP + CUDA). I also disabled the MapInterface because we need to replace the parsing library. We use Boost to parse but one of the function that is called during parsing is not compatible with CUDA 10 (https://stackoverflow.com/questions/55143135/error-cannot-call-member-function-impl-of-impl-pointer-in-boost)

.jenkins
```diff
@@ -0,0 +1,57 @@
...
+                sh '''
+                  cmake \
+                  -D BUILD_SHARED_LIBS=ON \
+                  -D CMAKE_BUILD_TYPE=Debug \
+                  -D DataTransferKit_ENABLE_DataTransferKit=ON \
+                  -D DataTransferKit_ENABLE_TESTS=ON \
+                  -D TPL_Trilinos_DIR=${TRILINOS_DIR} \
+                  -D TPL_BoostOrg_INCLUDE_DIRS=${BOOST_DIR}/include \
+                  -D CMAKE_CXX_EXTENSIONS=OFF \
+                  -D DataTransferKit_CXX11_FLAGS="-std=c++14 --expt-extended-lambda" \
```
dalg24: `--expt-extended-lambda` flag is not exported correctly by Kokkos/Trilinos?

Rombur: No :( I had to add it myself

dalg24:
You need to disable the old CI.  LGTM otherwise

PR #570, opened 2020-07-14
Update arborx

Rombur:
This PR works (I've only tried CPU) if I revert https://github.com/arborx/ArborX/pull/222 @masterleinad can you take a look at what we need to make `sendAcrossNetwork` work. One problem is obviously [this](https://github.com/arborx/ArborX/pull/222#pullrequestreview-431922791) but when I fixed it. I got the wrong results for the `MeshFree` tests.

### DIFF
--- a/packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
+++ b/packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
@@ -57,14 +57,16 @@ struct Helper
                                         Teuchos::FancyOStream &out )
     {
         ArborX::Details::Distributor<DeviceType> distributor( comm );
-        distributor.createFromSends( ranks );
+        distributor.createFromSends( typename DeviceType::execution_space{},
+                                     ranks );
 
         // NOTE here we assume that the reference solution is sized properly
         auto v_imp =
             Kokkos::create_mirror( typename View2::memory_space(), v_ref );
 
-        ArborX::Details::DistributedSearchTreeImpl<
-            DeviceType>::sendAcrossNetwork( distributor, v_exp, v_imp );
+        ArborX::Details::DistributedSearchTreeImpl<DeviceType>::
+            sendAcrossNetwork( typename DeviceType::execution_space{},
+                               distributor, v_exp, v_imp );
 
         // FIXME not sure why I need that guy but I do get a bus error when it
         // is not here...
@@ -110,7 +112,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DetailsDistributedSearchTreeImpl,
     Kokkos::fence();
 
     Kokkos::View<int *, DeviceType> ranks_u( "", comm_size );
-    ArborX::iota( ranks_u, 0 );
+    ArborX::iota( ExecutionSpace{}, ranks_u, 0 );
 
     Kokkos::View<int **, DeviceType> u_ref( "u_ref", comm_size, DIM );
     Kokkos::parallel_for( Kokkos::RangePolicy<ExecutionSpace>( 0, comm_size ),
@@ -186,7 +188,7 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DetailsNearestNeighborOperatorImpl, fetch,
 
     // v(i) <-- k*comm_size+i (index i, rank k)
     Kokkos::View<int *, DeviceType> v_exp( "v", comm_size );
-    ArborX::iota( v_exp, comm_rank * comm_size );
+    ArborX::iota( ExecutionSpace{}, v_exp, comm_rank * comm_size );
 
     Kokkos::View<int *, DeviceType> v_ref( "v_ref", comm_size );
     Kokkos::parallel_for( Kokkos::RangePolicy<ExecutionSpace>( 0, comm_size ),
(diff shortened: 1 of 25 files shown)


Rombur:
This is ready for review. The travis fails because of doxygen in arborx.

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
```diff
@@ -123,44 +125,51 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( DetailsDistributedSearchTreeImpl,
...
+    //        Kokkos::RangePolicy<ExecutionSpace>( 0, comm_size * comm_rank ),
+    //        KOKKOS_LAMBDA( int i ) {
+    //            for ( int j = 0; j < DIM; ++j )
+    //                v_ref( i, j ) = j * comm_rank;
+    //        } );
+    //    Kokkos::fence();
+    //
+    //    Helper<DeviceType>::checkSendAcrossNetwork( comm, ranks_v, v_exp,
+    //    v_ref,
+    //                                                success, out );
```
masterleinad, line 172: Why did you comment all of this?

packages/Meshfree/test/tstMovingLeastSquaresOperatorSimpleProblem.cpp
```diff
@@ -113,38 +117,12 @@ void checkResults( std::vector<double> const &values,
 TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( MovingLeastSquaresOperatorSimpleProblem,
                                    corner_cases, DeviceType )
 {
-    // No target point and no source point.
-    {
-        std::vector<std::array<double, DIM>> source_points = {};
```
masterleinad, line 118: Why did you delete this?

packages/Meshfree/test/tstNearestNeighborOperator.cpp
```diff
@@ -112,20 +107,8 @@ TEUCHOS_UNIT_TEST_TEMPLATE_1_DECL( NearestNeighborOperator, unique_source_point,
     DataTransferKit::NearestNeighborOperator<DeviceType> nnop(
         comm, source_points, target_points );
 
-    Kokkos::View<double *, DeviceType> source_values( "in" );
-    Kokkos::View<double *, DeviceType> target_values( "out" );
-
-    // violate pre condition of apply (source not properly sized)
```
masterleinad, line 118: ... and this

packages/Meshfree/test/tstDetailsCommunicationHelpers.cpp
Rombur: That's when I was debugging my problem with `sendAcrossNetwork`. It should not be commented.

packages/Meshfree/test/tstMovingLeastSquaresOperatorSimpleProblem.cpp
Rombur: For two of the cases the code does not throw and the last one it throws inside Kokkos instead of DTK. Since the code still throws (from Kokkos) when you try call `apply`, I didn't really try to understand why it stopped throwing during the construction.

packages/Meshfree/test/tstNearestNeighborOperator.cpp
Rombur: Same reason as before: we don't throw a DTK exception anymore and I don't really care about it. The user just needs to be careful.

dalg24:
Can you comment to our relaxing the tolerance from 1e-14 to 1e-6

masterleinad:
> The travis fails because of doxygen in arborx.

We should probably just disable doxygen parsing arborx (either in this pull request or another one).

Rombur:
> Can you comment to our relaxing the tolerance from 1e-14 to 1e-6

Yeah, that's because we went from `double` to `float` in the point coordinates. I _think_ that the important computations are done on double but we need to cast from `float` to `double` at some point. I looked at a few checks were I've relaxed the tolerance and it was clear that it was just due to a different round-off (value of 2.500001 instead of 2.5).

dalg24:
> > Can you comment to our relaxing the tolerance from 1e-14 to 1e-6
> 
> Yeah, that's because we went from `double` to `float` in the point coordinates. I _think_ that the important computations are done on double but we need to cast from `float` to `double` at some point. I looked at a few checks were I've relaxed the tolerance and it was clear that it was just due to a different round-off (value of 2.500001 instead of 2.5).

Do you reckon we might have an issue here?  We could recompute the distance in dtk in principle.

Rombur:
> > > Can you comment to our relaxing the tolerance from 1e-14 to 1e-6
> > 
> > 
> > Yeah, that's because we went from `double` to `float` in the point coordinates. I _think_ that the important computations are done on double but we need to cast from `float` to `double` at some point. I looked at a few checks were I've relaxed the tolerance and it was clear that it was just due to a different round-off (value of 2.500001 instead of 2.5).
> 
> Do you reckon we might have an issue here? We could recompute the distance in dtk in principle.

In the current version of the code, this wouldn't help since **all** the points in dtk are using float now. I'll see how much work it is to use float only in arborx and double everywhere else.

Rombur:
OK I've switched back to double and it works using a tolerance of `1e-14` instead of `1e-6`. We get narrowing warning but that can be fixed in another PR

dalg24:
> OK I've switched back to double and it works using a tolerance of `1e-14` instead of `1e-6`. We get narrowing warning but that can be fixed in another PR

Great.  Tanks for fixing this.

PR #574, opened 2020-07-31
Fix node types in tests

masterleinad:
This should make sure that the templated tests are run again.


masterleinad:
I also included the changes from #576 here. It turns out that we need `DefaulHostExecutionSpace` instead `HostSpace` to make this work.

PR #595, opened 2021-02-09
Update build system and fix warnings

Rombur:
I was helping a user when I realized that we do not install our CMake configuration files. We only install DTK version files. This PR fixes that. Since I was changing the build system I updated Tribits. In newer versions of Tribits, we don't need to hack the C++11 flags to use C++14. So I fixed this at the same time. Finally, I fixed a bunch or warnings. 

Locally, I have test failing for this PR and for master. Let's see if this PR passes with our CI.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -37,8 +37,7 @@ pipeline {
                   -D TPL_Trilinos_DIR=${TRILINOS_DIR} \
                   -D TPL_BoostOrg_INCLUDE_DIRS=${BOOST_DIR}/include \
                   -D CMAKE_CXX_EXTENSIONS=OFF \
-                  -D DataTransferKit_CXX11_FLAGS="-std=c++14 --expt-extended-lambda" \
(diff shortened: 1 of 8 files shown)


Rombur:
Jenkins is not working. Probably the same problem that we had with MGMOL and adamantine a while back. I am working on it.

dalg24 (approved):

masterleinad (approved):
Looks reasonable.

Rombur:
This is ready to be merged. I had to do one modification since the reviews. I was using a `KOKKOS_LAMBDA` inside of a constructor which does not work with CUDA. So I created a new function to split the `IndexRange` pair. I put this function in a new file in `Utils`. We used to have this kind of functions (like `iota`) in the `Search` package but now this has been moved outside DTK that's why I decided to put that function in `Utils`.

masterleinad (approved):

PR #597, opened 2021-02-09
Remove sphinx from the Docker image

Rombur:
There is problem building the docker images because of sphinx. If I remember correctly, the reason to have sphinx in the image is for debugging purpose. So I've removed it. Let's see if that fixes the CI failure.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,11 +30,6 @@ RUN apt-get update && apt-get install -y \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


Rombur:
retest this please

dalg24:
Sphinx was used to generate the documentation

Rombur:
Right but the check is done using Travis.

dalg24:
The image was also used for development and add "everything" installed.  Anyway removing is fine.

dalg24 (approved):

PR #598, opened 2021-03-08
Update ArborX version

Rombur:
I have replaced `KokkosExt::ArithmeticTraits::epsilon<double>::value` with `DBL_EPSILON` because DTK had a dependency on `ArborX_DetailsKokkosExt.hpp` but that file doesn't exist anymore. Instead of depending on another file in `ArborX/Details`, I just use `DBL_EPSILON` directly.


dalg24 (approved):

masterleinad (approved):
Looks OK.

PR #599, opened 2021-03-15
Allow to have ArborX be a TPL

Rombur:
With this PR, I can use an external ArborX with DTK. You need to set the following configuration variables:
```
-D DataTransferKit_ENABLE_ArborX_TPL=ON
-D TPL_ArborX_DIR=/path/to/arborx 
```
The PR is pretty straightforward, the only thing is that when ArborX is a TPL I create a dummy `dtk_search` library.


aprokop (commented):
I'm a bit confused about proliferation of the var names. There are `DataTransferKit_ENABLE_ArborX_TPL`, `DataTransferKit_ARBORX_TPL`, `TPL_ENABLE_ArborX`, `ArborX_TPL` and `ARBORX_PACKAGE`. Are all of them necessary?

It could also be nice if we could aggregate all `if` statements across different files in a single place, something like
```cmake
    IF(${PACKAGE_NAME}_ENABLE_ArborX_TPL)
      SET(${PACKAGE_NAME}_ARBORX_TPL ON)
      SET(ARBORX_PACKAGE "ArborX")
      SET(ARBORX_TPL ArborX "cmake/TPLs/" SS)
      SET(TPL_ENABLE_ArborX "")
    ELSE()
      ...
    ENDIF()
`

Rombur:
>  Are all of them necessary?

They are necessary. Tribits is looking for these variables. At first, I had weird bugs where ArborX was a registered TPL at the beginning of the configuration (because one variable was set) but not at the end (because another one was not set).

dalg24:
I expect only `TPL_ENABLE_<TPL_NAME>` and `<TRIBITS_PACKAGE>_ENABLE_<TPL_NAME>` are required

dalg24 (approved):

aprokop (approved):

PR #600, opened 2021-07-09
Check less in one_topo_one_fe_three_dim_hdiv

masterleinad:
Fixes https://github.com/xsdk-project/xsdk-issues/issues/141 partly.
https://github.com/trilinos/Trilinos/commit/3bcdcb8 changed the scaling of the polynomial basis so the output of the tests also changes. A couple of lines we say 
> We set X = 1. I don't know what the FE looks like so this tests just check that we don't crash

Hence, I propose to only check that the sizes of the arrays are correct. Alternatively, we could run a fallback check if the result doesn't match the current reference values.


Rombur (approved):
This looks good. The HDIV was never really good anyway.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

PR #603, opened 2022-10-04
Updates for Trilinos 14

masterleinad:
Fixes https://github.com/xsdk-project/xsdk-issues/issues/175.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,8 @@ ELSE()
         FIND_PATH(_Kokkos_SETTINGS
             NAMES KokkosConfig.cmake
             PATHS ${KOKKOS_PREFIX}
-            ${TPL_Trilinos_LIBRARY_DIRS}/cmake/Kokkos)
+                  ${TPL_Trilinos_LIBRARY_DIRS}/cmake/Kokkos
+                  ${Trilinos_DIR}/../Kokkos)
         INCLUDE(${_Kokkos_SETTINGS}/KokkosConfig.cmake)
         ASSERT_DEFINED(Kokkos_DEVICES)
 
--- a/packages/Meshfree/src/DTK_DetailsSplineProlongationOperator.hpp
+++ b/packages/Meshfree/src/DTK_DetailsSplineProlongationOperator.hpp
@@ -43,11 +43,7 @@ class SplineProlongationOperator
     {
         // Create a range map.
         Teuchos::ArrayView<const GlobalOrdinal> domain_elements =
-#if TRILINOS_MAJOR_MINOR_VERSION >= 130500
             _domain_map->getLocalElementList();
-#else
-            _domain_map->getNodeElementList();
-#endif
         _lda = domain_elements.size();
 
         const auto old_size = domain_elements.size();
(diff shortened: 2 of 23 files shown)


packages/Meshfree/src/DTK_DetailsSplineProlongationOperator.hpp
```diff
@@ -97,8 +97,10 @@ class SplineProlongationOperator
         using DeviceType = typename Node::device_type;
         using ExecutionSpace = typename DeviceType::execution_space;
 
-        auto x_view = X.getLocalViewDevice();
-        auto y_view = Y.getLocalViewDevice();
+        Y.scale( beta );
+
+        auto x_view = X.getLocalViewDevice(Tpetra::Access::ReadOnly);
+        auto y_view = Y.getLocalViewDevice(Tpetra::Access::ReadWrite);
```
masterleinad, line 103:
```suggestion
        auto x_view = X.getLocalViewDevice(Tpetra::Access::ReadOnly);
        auto y_view = Y.getLocalViewDevice(Tpetra::Access::ReadWrite);
```

masterleinad:
Retest this please.

CMakeLists.txt
```diff
@@ -64,7 +64,8 @@ ELSE()
         FIND_PATH(_Kokkos_SETTINGS
             NAMES KokkosConfig.cmake
             PATHS ${KOKKOS_PREFIX}
-            ${TPL_Trilinos_LIBRARY_DIRS}/cmake/Kokkos)
+                  ${TPL_Trilinos_LIBRARY_DIRS}/cmake/Kokkos
+                  ${Trilinos_DIR}/../Kokkos)
```
dalg24, line 68: What is this 2nd path for?

dalg24 (commented):
Should we support only the latest Trilinos?
(don't care too much, just trying to understand your rational)

masterleinad:
> Should we support only the latest Trilinos?

Yes, I think it's sufficient to support the latest releases and I'm pretty sure this pull request is not backward-compatible.

dalg24:
> > Should we support only the latest Trilinos?
> 
> Yes, I think it's sufficient to support the latest releases and I'm pretty sure this pull request is not backward-compatible.

Then why guard for the version of Trilinos?

CMakeLists.txt
masterleinad: Good question. This was from the original pull request. Let's see if the CI is happy without it.

masterleinad:
> Then why guard for the version of Trilinos?

Ah, also from the original pull request. Fixed now.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

CMakeLists.txt
masterleinad: We still need it in the CI.

aprokop (commented):
LGTM. But a question: if we require just the new version of Trilinos, do we actually check that anywhere in CMake? I don't see any

masterleinad:
> LGTM. But a question: if we require just the new version of Trilinos, do we actually check that anywhere in CMake? I don't see any

I would just specify it in `spack`. I don't know if we ever checked the `Trilinos` version (or `ArborX` version) at configure time.

masterleinad:
I would appreciate it if we can get this merged so that we are ready for the `xSDK` release.

Rombur (approved):

PR #605, opened 2023-08-23
Find Kokkos manually to fix compiling when Trilinos uses external Kokkos

masterleinad:
Fixes https://github.com/xsdk-project/xsdk-issues/issues/214.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ ELSE()
     IF(${PACKAGE_NAME}_TRILINOS_TPL)
         LIST(APPEND CMAKE_PREFIX_PATH ${TPL_Trilinos_LIBRARY_DIRS}/.. /usr/local/cuda/lib64)
 
-        find_package(Kokkos REQUIRED QUIET)
+        find_package(Kokkos REQUIRED QUIET HINTS ${Trilinos_DIR}/../Kokkos)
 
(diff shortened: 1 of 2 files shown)


bartlettroscoe (commented):
Just curious why `KokkosConfig.cmake` would be manually included when `find_package(Kokkos ...)` is already being called?

CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ ELSE()
     IF(${PACKAGE_NAME}_TRILINOS_TPL)
         LIST(APPEND CMAKE_PREFIX_PATH ${TPL_Trilinos_LIBRARY_DIRS}/.. /usr/local/cuda/lib64)
 
-        find_package(Kokkos REQUIRED QUIET)
+        find_package(Kokkos REQUIRED QUIET HINTS ${Trilinos_DIR}/../Kokkos)
```
bartlettroscoe, line 63: This is not needed for the Spack build of DTK with older versions of Trilinos, is it?  Spack should be setting the Trilinos install dir in `CMAKE_PREFIX_PATH` where KokkosConfig.cmake will be found automatically.

masterleinad: Our CI which only sets `TPL_Trilinos_DIR` doesn't find `Kokkos`, though.

PR #606, opened 2023-10-13
Fix compiling with gcc-13

masterleinad:
Fixes https://github.com/xsdk-project/xsdk-issues/issues/233.

### DIFF
--- a/packages/Utils/src/DTK_ConfigDefs.hpp
+++ b/packages/Utils/src/DTK_ConfigDefs.hpp
@@ -21,13 +21,8 @@
 
 #include <string>
 
-namespace DataTransferKit
-{
-
 #include "DTK_Types.h"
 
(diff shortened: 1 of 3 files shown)


packages/Utils/src/DTK_ConfigDefs.hpp
```diff
@@ -21,13 +21,8 @@
 
 #include <string>
 
-namespace DataTransferKit
-{
-
```
masterleinad, line 26: Note that we were including `<cstdint>` inside `namespace DataTransferKit`.

aprokop (approved):
LGTM (pending CI)

packages/Utils/src/DTK_ConfigDefs.hpp
dalg24: Ugggg

